PR #14, opened 2022-11-07
fix `isort` imports

vict0rsch:


saleml:
Thanks
Do you know why they haven't been caught by pre-commit even though isort is in the pre-commit config ?

saleml (approved):

PR #24, opened 2023-03-28
(WIP) Support for generic environments

saleml:
**This PR is a work in progress**.

It aims at addressing the issue discussed [here](https://github.com/saleml/gfn/issues/18), by implementing some of the suggestions in the issue when applicable, or alternatives if found better.

Here is a rough outline of the planned changes as of today (28/03/2023) - The list will probably evolve as more features are implemented and ideas emerge:

**1. Environment and Containers**

- [x] Have an abstract `Actions` class, rather than requiring the action space to be discrete, and actions to be represented by integers from $0$ to $n_{actions} - 1$.
- [x] Incorporate the `Actions` subclass into the abstract `Env` class. The `step` function should take an `Actions` object as input rather than a `LongTensor` for example. Require that in addition to specifying a subclass of `States` via `make_States_class`, the user needs to define a `make_Actions_class` function.
- [x] Abstract away the masking (`forward_mask` and `backward_mask`) ~~using functions implemented in the `Actions` class that take as input a `States` object and return a compact representation of which actions are (or aren't) allowed at those states~~ In the current changed files, this is done via the `validate_actions` method of `Env`. `DiscreteEnv`s implement the `validate_actions` using the forward and backward masks.
- [x] Create a`DiscreteEnv` subclass of `Env` that alleviates the need of writing a `make_Actions_class` for each discrete environment. The `DiscreteEnv` writes it own generic `make_Actions_class` that needs to know the number of possible actions only. The goal is to allow easy backward compatibility. Ideally, existing environments wouldn't need to be modified, except maybe using `DiscreteEnv` rather than `Env` as a parent class.
- [x] Change the `Transitions` and `Trajectories` containers so that the `actions` attributes are `Actions` objects rather than `LongTensor`s.

**2. Modules Estimators**

- [x] Current `GFNModules` should output something more generic than `logits`, they could be arbitrary objects.
- [x] Remove GFNModules, and use `nn.Modules` exclusively
- [x] We need a more generic `PFEstimator` and `PBEstimator` of which  the current `LogitPFEstimator` and `LogitPBEstimator` would only be subclasses. These generic estimators could transform the output of the `GFNModules` to `torch.Distributions` objects. 
- [x] Change the current `LogitPFEstimator` and `LogitPBEstimator` to output `Categorical` distributions.

**3. Samplers**

- [x] Change the `ActionsSampler` to a generic one, where the estimator used at initialization is a probability estimator (returning a batched distribution for a batch of states), and implements a generic `sample` function.
  As a result of this change, there is no need to subclass the ActionsSampler anymore, but only use the right estimators for each problem. Even `BackwardActionsSampler` wouldn't be needed anymore, as it would suffice to use a `ProbabilityEstimator` on the parents (e.g. `DiscretePBEstimator` of `examples/estimators.py`)
- [x] Add PF and PB estimators to `examples/estimators.py` corresponding to the sampling done in the `Box` environment (adapt the code of the continuous gfn paper)
- [x] Go through `trajectories_sampler` and make the changes needed following those of the `ActionsSampler`
- [x] Make the tests in `test_samplers_and_trajectories.py` work
- [x] Add tests for samplers and trajectories in the `Box` environment, using both the estimators implemented above (Beta distributions...) and some simpler estimators such as uniform over children.
- ~~Change the `ActionsSampler` class so that it takes into account the estimator's `Distributions` outputs as inputs, and outputs `Actions` objects rather than `LongTensor`s in addition to the `actions_log_probs`.~~
- ~~Ensure the `DiscreteActionsSampler` subclass is not broken as a result of such a change.~~

**4. Losses**

- ~~[Ensure that the losses, as they are currently implemented, do not rely on the action distributions being categorical. Make the necessary changes.~~
- [x] Go through the losses and parametrizations, and adapt to changes in samplers.

**5. Testing**
- [x] Ensure that all unit tests still pass
- [x] Ensure that the codebase is able to reproduce the results published for those environments:
- - [x] HyperGrid (TB, DB, SubTB, FM)
- - [x] Box
- - [x] Add functional test that reproduces the results (within an acceptable range)

**6. Code optimization**
- [x] Optimization: https://github.com/saleml/torchgfn/issues/50

**7. Miscellaneous**

- [x] Add one or two simple environments (e.g. from the Continuous GFlowNets paper) to the repo, and ensure that the level of specification needed to implement them is satisfying (i.e. we only need to code what's really specific to the environment, rather than boilerplate code that could have been abstracted away in the changes above) 
- - [x] Box environment
- - [x] DAG environment
- - [x] Molecule environment

- [x] Make unit tests for all the new added components
- [x] Rename `maskless_step` into `step` and `step` to `masked_step` ?
- [x] Go through the TODOs in the codebase, decide what needs to be fixed soon
- [x] Remove useless parts of code, or those that are not core to the library (functioning of GFlowNets), and either remove them, or move them to `examples/`
- [x] Move `envs` outside `gfn` - Maybe `gfnenvs` ?
- [x] Rename to `gfntorch` or `torchgfn` ?
- [x] Make a release of the repo (e.g. v0.1) before merging this PR. This would allow backward compatibility with codebases that use the current version (or a slightly older one) of the code. I'm thinking mainly of the `GFN_vs_HVI` codebase and the `MIFlowNet` codebase.
- [x] Make this v0.1 release installable, via `pip`, where `pip install torchgfn` would install both `torchgfn`/`gfntorch` and `gfnenvs` ? or `[envs]`
- [x] Go through codebase, add docstrings where needed, including examples/tutorials
- [x] Merge the PR, and make a second release, call it v1, and make it installable  via pip
- [x] Scripts and examples should go into a separate folder, with their own `requirements.txt`
- [x] Manual GPU testing. All tests (including reproducing results) should work on GPU (solve all missing `device=...`)
- [x] Streamline poetry/pyproject/setup.py...
- [x] Continuous integration ! Tests when pushing to master, merging etc...
- [x] Go to the repos that use the first versions of this library, and make sure that all functionalities (e.g. sampling with temperature) exist in the current version

**8. Final tests**
- [x] Ensure that all unit tests still pass
- [x] Ensure that the codebase is able to reproduce the results published for the published environments


saleml:
@josephdviviano @vict0rsch , let me know if you'd like to discuss these changes in person, what you think of the plan, and if/where/when you think you could help.
@EricElmoznino If this is still of interest to you, I'd love to know your opinion on the plan as well (e.g. would some environments you're working with fit within the new suggested abstractions ?), and if you'd like to contribute to this PR.

josephdviviano:
Definitely want to help, this should be an interesting exercise. Will also be a good opportunity to seek to integrate some features present in other codebases if we so choose. I am most curious about how to handle `GFNModules` outputting arbitrary objects.

I do think a in person would help to orient the work, happy to Fri or next week :)

josephdviviano (approved):
lgtm

PR #26, opened 2023-03-29
DAG environment

tristandeleu:
A WIP implementation of the DAG environment from DAG-GFlowNet.

### DIFF
--- a/src/gfn/envs/dag.py
+++ b/src/gfn/envs/dag.py
@@ -1,0 +1,185 @@
+from typing import ClassVar, Literal, Tuple
+
+import torch
+import numpy as np
+from gymnasium.spaces import Discrete
+from torchtyping import TensorType
+from itertools import product, permutations
+
+from gfn.containers.states import States
+from gfn.envs.env import Env
+
+# https://oeis.org/A003024
+NUM_DAGS = [1, 1, 3, 25, 543, 29281]
+
+# Typing
+TensorLong = TensorType["batch_shape", torch.long]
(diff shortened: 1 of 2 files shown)


src/gfn/envs/dag.py
```diff
@@ -0,0 +1,185 @@
...
+                    self.batch_shape + (env.num_variables ** 2,),
+                    **kwargs
+                )
+                return (forward_masks, backward_masks)
+            
+            def update_masks(self):
+                # TODO: The update of the transitive closure of the graphs
+                # (required to get the forward masks) depends on the last action
+                # applied. Reference: https://github.com/tristandeleu/jax-dag-gflownet/blob/53350bbfba3ab24f9ffa8196a3de42f139d9c5cc/dag_gflownet/env.py#L112-L116
+                self._closure_T = self._closure_T
```
tristandeleu, line 103: @saleml The update of the tensor necessary for computing the forward masks (`_closure_T`) depends on the last action applied (the one that led to the current state). We could implement it by keeping track of the last state inside the state object and reconstructing the action this way, but there may be a cleaner way to do that.

saleml: Thanks, I'm looking into it now

josephdviviano:
@tristandeleu I know this is quite old work but I would like to work on getting it into `torchgfn`. Has there been further progress on the method / environment side that I should integrate into this PR?

I need to do some work to align this with our new streamlined API in any case.

Thanks!

saleml:
closing as #296 implements this

PR #28, opened 2023-04-12
Env discrete states bugfix

josephdviviano:
Valid actions now properly indexes states AND actions, before checking if they are valid.

Trying to submit a small PR but this inherits a lot of our previous work. The only specific changes for this PR happen in 

`srv/gfn/envs/env.py`

in particular this tweak in `step` (changes starred):

```
    def step(
        self,
        states: States,
        actions: Actions,
    ) -> States:
        """Function that takes a batch of states and actions and returns a batch of next
        states and a boolean tensor indicating sink states in the new batch."""
        new_states = deepcopy(states)
        *valid_states_idx: TensorBool = ~states.is_sink_state
        *valid_actions = actions[valid_states_idx]
        *valid_states = states[valid_states_idx]

        if not self.validate_actions(valid_states, valid_actions):
            raise NonValidActionsError(
                "Some actions are not valid in the given states. See `is_action_valid`."
            )

```


saleml:
mergedi into https://github.com/saleml/gfn/pull/24

saleml (approved):

PR #29, opened 2023-04-13
States abstracted from containers

josephdviviano:
+ additional bugfixes for the `env.step()` and `env.backward_step()` methods.
+ `actions.py` and `states.py` moved outside of `containers/`.


saleml:
Merged into https://github.com/saleml/gfn/pull/24

PR #31, opened 2023-04-18
fixed imports

josephdviviano:
small number of import statement repairs

### DIFF
--- a/src/gfn/examples/__init__.py
+++ b/src/gfn/examples/__init__.py
@@ -1,0 +1,8 @@
+from .examples import (
+    DiscretePFEstimator,
+    DiscretePBEstimator,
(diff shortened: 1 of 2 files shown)


src/gfn/examples/__init__.py
```diff
@@ -0,0 +1,8 @@
+from .examples import (
```
saleml, line 1: `from .estimators` ?

saleml (changes requested):

src/gfn/examples/__init__.py
josephdviviano: thanks , I short circuited

saleml (approved):

PR #35, opened 2023-04-25
cleanup

josephdviviano:
Changes to the "annoying" parts of the package to streamline:

- removed `requirements.txt` and `setup.py` entirely - replaced with `pyproject.toml`.
- `flake8` dependency changed (moved away from gitlab which caused some errors on my side).
- `.pre.commit-config.yaml` dramatically simplified: now handles `black`, `flake8`, and `pytest`.
- `README.md` reformatted to reflect new install. Note one can now instantiate three envs:
    - `pip install .` : basic environment.
    - `pip install .[wandb]` : basic environment with training tools.
    - `pip install .[dev,wandb]` : full development environement with training tools.
- `.github/workflows/ci.yaml`: template continuous integration tests (I need help or permissions to setup actions.


vict0rsch:
I would add `isort` to `pre-commit` it's nice to have imports sorted and consistent

josephdviviano:
no good reason - I wasn't clear on how important that one was -- I'll add it back!

josephdviviano:
Added back `pyright` and  `isort`. Note that `isort` is causing a circular dependency in one of the tests I still need to figure out.

josephdviviano:
Note many new diffs due to import sorting.

pyproject.toml
```diff
@@ -1,7 +1,95 @@
...
+    "einops",
+    "gymnasium",
+    "numpy",
+    "torch",
+    "torchtyping",
+]
 
+
+[project.optional-dependencies]
+wandb = [
```
saleml, line 46: IIUC, naming this `wandb` means that `pip install .[wandb]` installs the extra dependencies. Wouldn't it make more sense then to call it `scripts` or something like this ?

README.md
```diff
@@ -53,7 +50,7 @@ python train.py --env HyperGrid --env.ndim 4 --env.height 8 --env.R0 0.01 --loss
 
 ### Example, in a few lines
 
-(⬇️ This example require the [`tqdm`](https://github.com/tqdm/tqdm) package to run. `pip install tqdm` or install all extra requirements with `pip install -r requirements.txt`)
+This example requires [`tqdm`](https://github.com/tqdm/tqdm) package to run. Use `pip install tqdm` or install all extra requirements with `pip install .[wandb]`.
```
saleml, line 53: See comment in `pyproject.toml`

pyproject.toml
josephdviviano: yeah that makes sense - I called it scripts so people don't think they are only installing `wandb`

PR #36, opened 2023-04-28
Continuous probability estimators

saleml:
This draft PR paves the way for estimators that can be used in continuous environments. It adapts the distributions used in [the continuous gfn paper's Box experiments](https://github.com/saleml/continuous-gfn) to the chosen abstractions.

I would like your opinion this, given that it's an example of what users would need to implement if they use their own (non-discrete environment). Basically, subclassing `torch.distributions.Distribution`, by writing `rsample` and `log_prob` functions, is the way to go here. The goal is to make the `ActionsSampler` agnostic to the estimator, as long as it outputs a probability distribution with these two methods.


saleml:
Also note the `UnsqueezedCategorical` as an alternative to the regular `Categorical` given that Discrete environments require unsqueezed actions. I need to double check whether `log_prob` needs to be rewritten for this class as well.

src/gfn/examples/distributions.py
```diff
@@ -0,0 +1,221 @@
+import numpy as np
+import torch
+from torch.distributions import Beta, Categorical, Distribution, MixtureSameFamily
+from torchtyping import TensorType
+
+
+class UnsqueezedCategorical(Categorical):
+    """This class is used to sample actions from a categorical distribution. The samples
+    are unsqueezed to be of shape (batch_size, 1) instead of (batch_size,).
```
josephdviviano, line 9: comment as to why?

src/gfn/examples/distributions.py
```diff
@@ -0,0 +1,221 @@
...
+        assert beta.shape == (self.n_states, self.n_components)
+
+        self.base_dist = MixtureSameFamily(
+            Categorical(logits=mixture_logits),
+            Beta(alpha, beta),
+        )
+
+        self.min_angles, self.max_angles = self.get_min_and_max_angles()
+
+    def get_min_and_max_angles(self):
```
josephdviviano, line 54: General comment - the code is nicely self-documenting but I think it would be good for us to go through at some point and ensure every method has some minimum level of docstring.

src/gfn/examples/distributions.py
```diff
@@ -0,0 +1,221 @@
...
+            )
+            * self.delta
+        )
+
+        if self.northeastern:
+            sampled_next_states = self.centers + sampled_actions
+        else:
+            sampled_next_states = self.centers - sampled_actions
+        # TODO: do we need to return actions or next states here? -- logprobs are ok, but the previous 4 lines might not be
+        return sampled_next_states
```
josephdviviano, line 101: I don't fully understand this TODO -- my immediate thought if that `sampled_next_states`  should *always* be the only thing we return ?

src/gfn/examples/distributions.py
```diff
@@ -0,0 +1,221 @@
...
+        base_01_logprobs = self.base_dist.log_prob(base_01_samples)
+
+        logprobs = (
+            base_01_logprobs
+            - np.log(self.delta)
+            - np.log(np.pi / 2)
+            - torch.log(self.max_angles - self.min_angles)
+        )
+
+        return logprobs
```
josephdviviano, line 126: I find these two implementations extremely clean. For this example we could comment liberally or make explicit the math in the form of equations - but I like the work flow of `next_states = obj.sample()` -> `logprobs = obj.calc_logprob(next_states)`.

src/gfn/examples/estimators.py
```diff
@@ -78,4 +79,16 @@ def to_probability_distribution(
         logits = module_output
         logits[~states.backward_masks] = -float("inf")
 
-        return Categorical(logits=logits)
+        return UnsqueezedCategorical(logits=logits)
```
josephdviviano, line 82: Will we ever need `SqueezedCategorical` is my question -- if we need the unsqueeze for broadcasting, I think we could just overwrite this name `Categorical` from `torch.distributions` with the one specific to `gfn` so that the user never accidentally uses the vanilla `Categorical`?

josephdviviano (approved):
Looks nice - I like the `rsample` and `log_probs` workflow, and the reliance on distributions is quite clean.

As an aside I actually like having `TensorType[]` explicitly defined in each signature instead of centralizing typing and having to remember which is which.

I think we should go for either hard centralization as in my PR or defining them explicitly everywhere as in your example here -- not a mix-and-match approach. I'm open minded, let me know what you think.

saleml:
> As an aside I actually like having `TensorType[]` explicitly defined in each signature instead of centralizing typing and having to remember which is which.
> 
> I think we should go for either hard centralization as in my PR or defining them explicitly everywhere as in your example here -- not a mix-and-match approach. I'm open minded, let me know what you think.

Of course, we should not have mix-and-match. The reason I was explicit in the signatures here is to actually get your opinion on this way of typing.

With centralized typing, as in your PR, you are right that we need to remember which type does what. See the following screenshot for example when I hover the type.

Screenshot 2023-05-01 at 6 23 21 PM

I am slightly in favor of explicit typing. We could make the signatures shorter with something like `from torchtyping import TensorType as T`, and choosing shorter (but consistent) names for the string dimensions.

If we go this route, it would cancel some of the work you did in that PR though :/ 

Thoughts @vict0rsch on the two options ?

src/gfn/examples/distributions.py
saleml: Actions are useful when printing trajectories. But it probably is their only benefit. If we decide we do not care about this, we could just have action samplers sample next states, and discard actions altogether.

src/gfn/examples/distributions.py
saleml: Right, there is some math in the paper justifying this code. I'll incorporate some of it in the comments.

josephdviviano:
@saleml I don't mind if the work is cancelled in the other PR - what matters is the final product is easy to use :) 

@vict0rsch would be happy to hear your thoughts!

For context I left some pros/cons here https://github.com/saleml/gfn/pull/38#issuecomment-1530123504 - I'd like to handle the typing via that PR - but am now leaning also towards explicitly writing types for all signatures. If we all vote for that I'll do the work asap (and rename the other PR if I need to :P ).

src/gfn/examples/estimators.py
saleml: I'm not sure what you mean by "overwriting the name", do you mean just rename `UnsqueezedCategorical` to `Categorical` ? 
Btw the user should *never* have to use Categorical or UnsqueezedCategorical, for discrete environments, given that the distribution is abstracted away when using `DiscretePFEstimator` or `DiscretePBEstimator` (as in the README example). 
The user would need to write their own distributions for custom environments only (as what I'm doing in this PR [although not completely done] with `Box`)

josephdviviano: Yes that's what I meant by "overwriting". I was assuming that a major use case of the lib would be for people to implement their own envs (since that's going to be the most domain-specific implementation piece) but I also take your point that the specific distributions are going to be based on the specific environments, so we shouldn't automagically overwrite any of the distribution classes.

saleml:
This is ready for review. What do you think of having the file `box_utils.py` under `examples` ? Basically, these classes are examples of what the user would need to write for their specific environment. 
I chose to put them under `examples` because they're not core to the library per se, and are only used for the `Box` environment.

An alternative would be to write the estimators in `utils/estimators.py` and distributions in `utils/distributions.py`, but I find the estimators and distributions of this file to be too specific for one environment to be actual "utils".

scripts/configs/sampler.py
saleml, line 1: to disregard for now - configs will be the last thing we'll modify

src/gfn/losses/detailed_balance.py
saleml, line 1: disregard for now, we'll do losses afterwards

src/gfn/losses/flow_matching.py
saleml, line 1: same

src/gfn/losses/sub_trajectory_balance.py
saleml, line 1: same

src/gfn/losses/trajectory_balance.py
saleml, line 1: same

saleml:
Adding tests right now

vict0rsch:
Guys, I totally missed the pings. Don't hesitate to shake the tree on slack if that happens again :p I'll have a deeper look tomorrow

src/gfn/examples/box_utils.py
```diff
@@ -0,0 +1,457 @@
...
+                northeastern=True,
+                centers=states,
+                exit_probability=exit_probability,
+                mixture_logits=mixture_logits,
+                alpha=alpha,
+                beta=beta,
+            )
+
+
+class BoxPBEstimator(ProbabilityEstimator):
```
josephdviviano, line 323: I'm skeptical about the requirement to define an environment specific estimator. I feel like this shouldn't be required if we had the correct abstraction. I wonder what @vict0rsch thinks?

josephdviviano (approved):

josephdviviano:
I like this PR a lot but I'd like to push back on swapping `setuptools` for `poetry`, and I'm curious about the BoxEstimator classes you define - I'm not clear on why they're needed.

src/gfn/actions.py
```diff
@@ -106,8 +106,8 @@ def compare(
         Returns:
             boolean tensor of shape batch_shape indicating whether the actions are equal
         """
-        out = self.tensor == other.tensor
-        n_batch_dims = len(self.__class__.batch_shape)
```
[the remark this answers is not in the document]
saleml: The logic I followed so far is:
- `self.x` to access an instance variable
- `self.__class__.x` to access a class variable (i.e. x is the same for all instances of the class)

src/gfn/examples/box_utils.py
```diff
@@ -0,0 +1,457 @@
...
+            return QuarterDisk(
+                delta=self.env.delta,
+                mixture_logits=mixture_logits,
+                alpha_r=alpha_r,
+                beta_r=beta_r,
+                alpha_theta=alpha_theta,
+                beta_theta=beta_theta,
+            )
+        else:
+            # we check that the module_output is of shape (*batch_shape, 1 + 3 * n_components)
```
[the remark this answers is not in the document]
saleml: ok done

pyproject.toml
```diff
@@ -1,66 +1,61 @@
 [build-system]
-requires = ["setuptools>=61.0"]
```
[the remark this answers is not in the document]
saleml: When doing the release to pypi last week (master branch), @vict0rsch  made the switch from setuptools to poetry. I forgot the reason. @vict0rsch, care to remind us ?
I figured we might just as well do the switch now in generic-gfn, rather than having to deal with both whenever we switch branches.
Things work nicely with poetry however.

src/gfn/actions.py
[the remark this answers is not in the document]
josephdviviano: thanks that makes sense

pyproject.toml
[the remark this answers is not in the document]
josephdviviano: OK if things work well with poetry when pip installing from `pypy` I have no qualms :)

src/gfn/examples/box_utils.py
saleml: Let me summarize the expected workflow with the current codebase, and then please let me know what abstractions we can change to make it better:
Given an environment, how do we sample trajectories ?
- We need a distribution over actions that is conditioned on the state. For discrete environments for example, the distribution is a Categorical one. It can be anything we want for other environments. So there is always **a need for a `torch.Distribution`**.
- - Example: For the Box environment, in this PR, I coded Mixture of Betas (scaled). Why ? The environment is bounded, so we need a parametrized distribution that gives us actions within a bound. Beta's sample space is [0, 1], but is usually unimodal. Thus the mixture, and the scaling.
- This distribution, whether it's categorical or Beta, or whatever, has some parameters $\theta(s)$. For example, for Beta, it's the concentration parameters $\alpha$ and $\beta$.
- $\theta(s)$ is usually a neural network. So there is, almost always, **a need for a `nn.Module`, that takes as input a batch of states, and returns the parameters of the distribution**
- The `PFEstimator` and `PBEstimator` serve just one purpose: **transforming the output of the `nn.Module`, to the `Distribution`**, eventually by applying some modifications (please see the new PR for an updated version of this file, where there are some modifications - the scaling -, and there is a Neural Net implemented).

An alternative that is aligned with your suggestion (i.e. not asking the user to code all 3 of: `Distribution`, `nn.Module`, `PF/BEstimator`), is to do the postprocessing of `nn.Module`'s output, in the `nn.Module`. So we would need some abstract `GFNModule` class that subclasses `nn.Module` but requires the implementation of `self.to_probaiblity_distribution(...)` that is now in the `ProbabilityEstimator` class.

Thoughts ?

PR #39, opened 2023-05-02
Tutorials for `gfn`

vict0rsch:


saleml:
Looking at how the `examples` folder is used here (notebooks, and scripts), I think it makes more sense to move `examples/[modules,estimators,distributions].py` to `utils/[...].py`. 
The classes in the three files are neither examples per se nor core part of the library, but utilitaries that may be helpful for many people.

Thoughts, @vict0rsch @josephdviviano ?

josephdviviano:
Yeah that makes sense. `examples` could be more scripts that stitch all elements together, similar to the example in the `README.md`.

saleml:
@vict0rsch Can you change where the branch will be merged to (`master`), so that it can be reopened ?

saleml:
reopen

PR #40, opened 2023-05-03
Better example in README + examples to utils

saleml:
Following https://github.com/saleml/gfn/pull/39 and the slack discussion. This PR serves two purposes:
- Rename examples folder to utils, so that examples is exclusively used for scripts and notebooks
- Provide a (close to?) working example in the README that incorporates the recent API changes. This should help  @vict0rsch design better tutorials and notebooks.


src/gfn/utils/estimators.py
```diff
@@ -32,7 +34,7 @@ def __init__(
             sf_bias (float, optional): scalar to subtract from the exit action logit before dividing by temperature. Defaults to 0.0.
             epsilon (float, optional): with probability epsilon, a random action is chosen. Defaults to 0.0.
         """
-        super().__init__(**kwargs)
+        super().__init__(env, module)
```
josephdviviano, line 37: nice!

josephdviviano (approved):
thanks for the PR - even if most of the time we just accept with no or minimal changes, it's helpful for keeping everyone on the same page.

vict0rsch (approved):

josephdviviano:
I'm thinking it makes sense to have both (only one needs to be in the
actual readme) to show the user how the different elements of the codebase
interact under a few conditions (these losses are different enough that you
have to change multiple elements, and that diff highlights to the user how
the pieces work together)

Joseph Viviano
viviano.ca

PR #41, opened 2023-05-05
Small fixes

vict0rsch:
I initially just wanted to make `__version__` available.

Then I fixed a few more things

* `isort` to sort imports consistently
* Bring back `gfn.utils.modules.Uniform` from the legacy `ZeroGFNModule`
	* might not be right, but it's used in `dynamic_programming.py` so do update as you see fit @saleml 
* Rename `utils.py` to `utils/common.py` because `utils.py` and `utils/` can't really coexist. You could put the contents of `common.py` into `utils/__init__.py` if you want to keep the `from gfn.utils import X`


josephdviviano (approved):
Nice! How did you fix `isort`?

src/gfn/utils/modules.py
```diff
@@ -87,3 +87,19 @@ def __call__(self, preprocessed_states: InputTensor) -> OutputTensor:
         assert preprocessed_states.dtype == torch.long
         outputs = self.table[preprocessed_states.squeeze(-1)]
         return outputs
+
+
+class Uniform(nn.Module):
+    def __init__(self, output_dim: int) -> None:
+        """Implements a zero function approximator, i.e. a function that always outputs 0.
```
saleml, line 94: The reason I had removed it is that it doesn't make sense to have a uniform distribution (over parents or children) for a generic environment. But I agree it would be useful to have it for discrete environments. Could you make a better docstring, mentioning e.g. that output_dim is usually n_actions - 1 in discrete environments (representing the maximum number of parents)

saleml: @vict0rsch ?

vict0rsch:
> Nice! How did you fix `isort`?

I didn't include it in pre-commit, rather just ran it manually

PR #47, opened 2023-05-30
Explicit typing

josephdviviano:
All function signatures now explicitly define the `TensorType` (imported as `TT` for cleanliness) in the signature:

E.g.,

```
class Loss(ABC):
    "Abstract Base Class for all GFN Losses"

    def __init__(self, parametrization: Parametrization):
        self.parametrization = parametrization

    @abstractmethod
    def __call__(self, *args, **kwargs) -> TT[0, float]:
        pass
```

or in a more complex example:

```
class Trajectories(Container):
    def __init__(
        self,
        env: Env,
        states: States | None = None,
        actions: Actions | None = None,
        when_is_done: TT["n_trajectories", torch.long] | None = None,
        is_backward: bool = False,
        log_rewards: TT["n_trajectories", torch.float] | None = None,
        log_probs: TT["max_length", "n_trajectories", torch.float]
        | None = None,
    ) -> None:
    ...
```

Lots of diffs, but that's it.


src/gfn/actions.py
```diff
@@ -99,37 +91,37 @@ def extend_with_dummy_actions(self, required_first_dim: int) -> None:
...
         Args:
             other: tensor of actions
         Returns:
             boolean tensor of shape batch_shape indicating whether the actions are equal
         """
-        out = self.actions_tensor == other
+        out = self.tensor == other.tensor
         action_ndim = len(self.__class__.action_shape)
-        for _ in range(action_ndim):
+        for _ in range(action_ndim):  # TODO: this seems over complex?
```
saleml, line 111: do you know of a simpler alternative ? It felt weird writing these lines indeed

src/gfn/examples/distributions.py
saleml, line 1: could you remove this file from the PR ? It's not necessary and will just make merging https://github.com/saleml/torchgfn/pull/36 harder

saleml (commented):
Great ! Thanks for the PR.

src/gfn/actions.py
josephdviviano: looking at it now!

josephdviviano: Check this out - we can flatten all actions dims assuming the batch dims are always defined: https://github.com/saleml/torchgfn/pull/47/commits/3a85dd9cebdc80a2160ee8125484f438204655f6

saleml: Awesome !

PR #55, opened 2023-06-08
Adapt parametrizations and losses to new API

saleml:
All tests pass, except those related to parametrization/loss of the Box environment (we need the fancy Distribution wrapper for that)


src/gfn/actions.py
```diff
@@ -34,6 +34,10 @@ def __init__(self, tensor: TT["batch_shape", "action_shape"]):
             tensor: tensor of actions
         """
         self.tensor = tensor
+        assert len(tensor.shape) > len(self.action_shape), (
```
josephdviviano, line 37: I know it's pedantic but I'd add the comment

`# Ensure the tensor has all action dimensions and at least one batch dimension.`

src/gfn/casting.py
```diff
@@ -1,29 +0,0 @@
-# TODO: remove this file eventually
```
josephdviviano, line 1: you love to see it

src/gfn/examples/box_utils.py
```diff
@@ -377,6 +377,19 @@ def forward(
...
+    """A module to be used to create a uniform PB distribution for the Box environment
+
+    A module that returns (1, 1, 1) for all states. Used with QuarterCircle, it leads to a
+    uniform distribution over parents in the south-western part of circle.
+    """
+
+    def forward(
+        self, preprocessed_states: TT["batch_shape", 2, float]
+    ) -> TT["batch_shape", 3]:
+        return torch.ones(preprocessed_states.shape[:-1] + (3,))
```
josephdviviano, line 390: Worth a comment explaining the shape here (I know you also have it elsewhere, but I'm being annoying).

josephdviviano (approved):
In the interest of time I didn't carefully review the tests but the changes to the main code are really nice. Seems like you managed to mostly make the code shrink, which is a huge plus.

PR #56, opened 2023-06-09
💥💥

saleml:
💥 No more redundant `Loss` classes. Losses are evaluated directly from the parametrizations (need to find better names).

💥 Parametrizations now have a `sample_trajectories` method, that calls the `ActionsSampler` and `TrajectoriesSampler` under the hood. The user does not need to create these anymore, except maybe for fancy sampling (off-policy e.g.). Still room for improvement for this one: should we keep the `Sampler` classes anyway ? 

💥 New script `train_hypergrid.py`, that doesn't need any config file / simple parsing. Self-contained to allow for all possible losses. The script also shows some commands on the top that make it possible to **reproduce published results** in this environment. (Except for the flow matching loss, which is for some reason more efficient in terms of learning curves in this library - I need to double check with paper authors who trained with the flow matching loss, to decide whether our results are comparable or not)

💥 All tests pass except those related to the `Box` environment in `test_parametrizations_and_losses.py`, as they need a component @josephdviviano is working on

For discrete environments, the `train_hypergrid.py` script is supposed to show what a user needs to write, in addition to their environment.

The nested `if`s are a bit ugly. `simple_parsing` and `pyyaml` are handling that more elgantly in the old (to be removed ?) script `train.py`. We need to decide what script(s) to leave in the repo.

Obligatory emojis in the title of this PR because we're almost there wrt the checklist in https://github.com/saleml/torchgfn/pull/24. Just a few tweaks left !


vict0rsch:
Will do tomorrow. Looks good

src/gfn/containers/replay_buffer.py
josephdviviano, line 1: beautiful!

src/gfn/losses/flow_matching.py
```diff
@@ -20,8 +20,19 @@ class FMParametrization(Parametrization):
...
+
+    Attributes:
+        logF: LogEdgeFlowEstimator
+        alpha: weight for the reward matching loss.
     """
     logF: LogEdgeFlowEstimator
+    alpha: float = 1.0
+
+    def __post_init__(self) -> None:
+        self.env = self.logF.env
```
josephdviviano, line 35: Just another comment with no action required - I'm definitely going to try to figure out how to *not* pass the `env` everywhere in a follow up PR after the release.

src/gfn/losses/sub_trajectory_balance.py
```diff
@@ -7,87 +7,55 @@
...
-        self.log_reward_clip_min = log_reward_clip_min
-        self.weighing = weighing
-        # Lamda is a discount factor for longer trajectories. The part of the loss
-        # corresponding to sub-trajectories of length i is multiplied by lamda^i
-        # where an edge is of length 1. As lamda approaches 1, each loss becomes equally
-        # weighted.
-        self.lamda = lamda
+    logF: LogStateFlowEstimator
+    on_policy: bool = False
+    weighing: Literal[
```
josephdviviano, line 48: This is the right place to check whether arguments passed are correct.

josephdviviano (approved):
Minor comments - feel free to ignore as they can easily be picked up in future PRs.

scripts/train_hypergrid.py
```diff
@@ -0,0 +1,234 @@
...
+    DBParametrization,
+    FMParametrization,
+    LogPartitionVarianceParametrization,
+    SubTBParametrization,
+    TBParametrization,
+)
+from gfn.utils.common import trajectories_to_training_samples, validate
+from gfn.utils.estimators import DiscretePBEstimator, DiscretePFEstimator
+from gfn.utils.modules import DiscreteUniform, NeuralNet, Tabular
+
```
vict0rsch, line 32:
```
if __name__ == "__main__":
```

scripts/train_hypergrid.py
```diff
@@ -0,0 +1,234 @@
...
+parser = ArgumentParser()
+
+parser.add_argument("--no_cuda", action="store_true")
+
```
vict0rsch, line 36: Add `help=` everywhere :)

scripts/train_hypergrid.py
```diff
@@ -0,0 +1,234 @@
...
+parser.add_argument("--validation_interval", type=int, default=100)
+parser.add_argument(
+    "--validation_samples",
+    type=int,
+    default=200000,
+    help="Number of validation samples to use to evaluate the pmf.",
+)
+
+parser.add_argument("--wandb", type=str, default="")
```
vict0rsch, line 88: `wandb_project` ?

scripts/train_hypergrid.py
```diff
@@ -0,0 +1,234 @@
...
+    wandb.config.update(args)
+
+# 1. Create the environment
+env = HyperGrid(
+    args.ndim, args.height, args.R0, args.R1, args.R2, device_str=device_str
+)
+
+# 2. Create the necessary modules, estimators, and parametrizations
+if args.loss == "FM":
+    # We need a LogEdgeFlowEstimator
```
vict0rsch, line 110: Why? (I mean: 2 words to explain)

scripts/train_hypergrid.py
```diff
@@ -0,0 +1,234 @@
...
+        module = NeuralNet(
+            input_dim=env.preprocessor.output_dim,
+            output_dim=env.n_actions,
+            hidden_dim=args.hidden_dim,
+            n_hidden_layers=args.n_hidden,
+        )
+    estimator = LogEdgeFlowEstimator(env=env, module=module)
+    parametrization = FMParametrization(estimator)
+else:
+    # We need a DiscretePFEstimator and a DiscretePBEstimator
```
vict0rsch, line 123: Same

scripts/train_hypergrid.py
```diff
@@ -0,0 +1,234 @@
...
+        "equal_within",
+    ],
+    default="geometric_within",
+)
+parser.add_argument("--subTB_lambda", type=float, default=0.9)
+
+parser.add_argument(
+    "--tabular", action="store_true", help="Use a lookup table for F, PF, PB"
+)
+parser.add_argument("--uniform", action="store_true", help="Use a uniform PB")
```
vict0rsch, line 68: `uniform_pb` would be much more explicit, including further in the code

scripts/train_hypergrid.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+visited_terminating_states = env.States.from_batch_shape((0,))
+
+states_visited = 0
+n_iterations = args.n_trajectories // args.batch_size
+for iteration in trange(n_iterations):
+    trajectories = parametrization.sample_trajectories(n_samples=args.batch_size)
+    training_samples = trajectories_to_training_samples(trajectories, parametrization)
+    training_objects = training_samples
```
vict0rsch, line 213: why do you need that?

src/gfn/losses/flow_matching.py
```diff
@@ -122,16 +107,19 @@ def flow_matching_loss(
     def reward_matching_loss(self, terminating_states: DiscreteStates) -> TT[0, float]:
         """Calculates the reward matching loss from the terminating states."""
         assert terminating_states.log_rewards is not None
-        log_edge_flows = self.parametrization.logF(terminating_states)
+        log_edge_flows = self.logF(terminating_states)
         terminating_log_edge_flows = log_edge_flows[:, -1]
         log_rewards = terminating_states.log_rewards
         return (terminating_log_edge_flows - log_rewards).pow(2).mean()
 
     # TODO: should intermediary_states and terminating_states be two input args?
```
vict0rsch, line 115: I would say yes IFF: (they can be decoupled) OR (they don't come from a single function/method call).

In other words, I would find arguments to be more idiomatic in those two situations:

```python
# decoupling
fm_param = FMParametrization(...)
some_states, term_states = some_function()
some_states = modify(some_states)
fm_param.loss(some_states, term_states) # better than fm_param.loss((some_states, term_states))

# multiple funcs
fm_param = FMParametrization(...)
some_states = some_function()
term_states = some_other_function()
fm_param.loss(some_states, term_states) # better than fm_param.loss((some_states, term_states))
```

src/gfn/utils/common.py
```diff
@@ -7,40 +7,39 @@
...
         return trajectories.to_non_initial_intermediary_and_terminating_states()
-    elif isinstance(loss_fn, TrajectoryDecomposableLoss):
+    elif isinstance(parametrization, TrajectoryDecomposableLoss):
         return trajectories
-    elif isinstance(loss_fn, EdgeDecomposableLoss):
+    elif isinstance(parametrization, DBParametrization):
         return trajectories.to_transitions()
     else:
-        raise ValueError(f"Loss {loss_fn} is not supported.")
+        raise ValueError(f"Parametrization {parametrization} not supported.")
```
vict0rsch, line 42: Another way would be to have 

```python
class Parametrization(...):
    ...

	@abstractmethod
    def trajectories_to_training_samples(self, trajectories):
        pass

class FMParametrization(...):
    ...

	def trajectories_to_training_samples(self, trajectories):
        return trajectories.to_non_initial_intermediary_and_terminating_states()
```

That's more OOP but lacks centralization. Food for thoughts

testing/test_parametrizations_and_losses.py
```diff
@@ -73,8 +62,11 @@ def test_FM(env_name: int, ndim: int, module_name: str):
 
     states_tuple = trajectories.to_non_initial_intermediary_and_terminating_states()
 
-    loss = FlowMatching(parametrization)
-    print(loss(states_tuple))
+    loss = parametrization.loss(states_tuple)
+    print(loss)
```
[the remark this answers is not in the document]
vict0rsch: I think it's reasonable to say that it should decrease regularly (with an earlystopping-like patience) and that at least it should not be `nan` or `inf`

vict0rsch (approved):

src/gfn/utils/common.py
saleml: Yes, I'll make an issue out of this

src/gfn/losses/flow_matching.py
saleml: I don't get what you're referring to here

saleml:
The last commits should address all your comments, except the one I didn't understand from @vict0rsch.

src/gfn/losses/flow_matching.py
josephdviviano: I'm also missing some context :)

josephdviviano:
Thanks - I'm happy to merge and address the follow up in a separate PR(s) if req so we don't block.

src/gfn/losses/flow_matching.py
vict0rsch: The comment reads "# TODO: should intermediary_states and terminating_states be two input args?" :p

PR #63, opened 2023-06-19
♻️ (losses): Sub-parameters separated by . not _

### LINKED ISSUE #62: logZ_logZ
let's fix this, and think of better ways of saving/loading parameterization's state dicts
In the latest branch: from https://github.com/saleml/torchgfn/blob/master/src/gfn/losses/base.py#L50 to the end of the class

vict0rsch:
Fix #62


vict0rsch:
Closing because of `master` base for `log_z[...]`

PR #69, opened 2023-06-30
Simple yaml configs (single-file version of `master`) + move envs to `examples/`

vict0rsch:
There's now an `examples/` folders that contains 

* `envs/` a folder with example envs, helper functions and a readme to explain envs in more details (TODO)
* `scripts` a folder with the training scripts and a readme to explain how to use them

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,7 @@ dev = [
     "sphinx_rtd_theme",
     "tox"
 ]
-scripts = ["tqdm", "wandb"]
+scripts = ["tqdm", "wandb", "scikit-learn", "scipy"]
 all = [
     "pre-commit",
     "pytest",
@@ -67,7 +67,9 @@ all = [
     "black",
     "myst-parser",
     "tqdm",
-    "wandb"
+    "wandb",
+    "scikit-learn",
+    "scipy"
 ]
 
 [project.urls]
--- a/src/gfn/actions.py
+++ b/src/gfn/actions.py
@@ -34,10 +34,10 @@ class Actions(ABC):
             tensor: tensor of actions
         """
         self.tensor = tensor
-        assert len(tensor.shape) > len(self.action_shape), (
+        assert len(tensor.shape) >= len(self.action_shape), (
             f"Actions tensor has shape {tensor.shape}, "
             f"but the action shape is {self.action_shape}."
-            # Ensure the tensor has all action dimensions and at least one batch dimension.
+            # Ensure the tensor has all action dimensions.
         )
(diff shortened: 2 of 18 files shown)


src/gfn/utils/misc.py
saleml, line 1: should it be in src ?

saleml:
Looks good to me. I left a question.

It might be simpler to merge #71 first

src/gfn/utils/misc.py
vict0rsch: what do you mean?

vict0rsch: Ah I understand ; the way I see it:

1. small cost
2. the point of this function is that wherever the "user" is, it will make `examples/` available to them. The "user" may be an example script, a test, or a human running the readme code sample. In this context, the only thing we can guarantee is that `gfn` will be in the path. The rest depends on where the user is and so putting `add_root_to_path` anywhere else than the source code would require an assumption of location

The alternative is to get rid of `add_root_to_path` and add `sys.path.append` calls to each file that requires it, i.e. adding

```
from pathlib import Path
import sys

sys.path.append(str(Path(__file__).resolve().parent.[...].parent))
```

saleml:
@vict0rsch move back

README.md
```diff
@@ -39,35 +39,33 @@ pip install .[scripts]
...
-from gfn.envs import HyperGrid
 from gfn.estimators import LogZEstimator
-from gfn.losses import TBParametrization, TrajectoryBalance
+from gfn.losses import TBParametrization
 from gfn.samplers import ActionsSampler, TrajectoriesSampler
-from gfn.utils import NeuralNet, DiscretePFEstimator, DiscretePBEstimator
+from gfn.utils import NeuralNet, DiscretePFEstimator, DiscretePBEstimator, add_root_to_path
+
+# just making sure the examples/ folder is available in the Python path
+add_root_to_path()
```
josephdviviano, line 59: I wonder whether this example should be an executable script you point to in the readme -- or, you can remove lines like this to keep this pseudocod-like, so the user can focus on the design pattern of the library and not have to worry about whether the code will actually run. In a readme the less lines of actual code the better, to highlight core design principles, I think.

examples/envs/Readme.md
josephdviviano, line 1: this is excellent!

src/gfn/utils/misc.py
josephdviviano: This comment is obsolete given what we discussed today right? Moving everything back into a single package?

josephdviviano (approved):

josephdviviano:
Looks great!

vict0rsch:
@saleml remove gymnasium dependency

PR #71, opened 2023-06-30
Make all tests related to Box and box_utils pass

saleml:
Screenshot 2023-06-30 at 7 01 51 PM

3 months later, tests pass


src/gfn/examples/box_utils.py
```diff
@@ -490,10 +510,16 @@ def forward(
             out_s0 = out_s0 * self.components_mask
 
         # Overwrite the network outputs with PFs0 in the case that the state is 0.0.
-        idx_s0 = torch.where(preprocessed_states == 0.0)[0]
+        idx_s0 = torch.all(preprocessed_states == 0.0, 1)
```
josephdviviano, line 513: woah that a nice trick

src/gfn/examples/box_utils.py
```diff
@@ -803,9 +830,8 @@ def to_probability_distribution(
             marker="x",
         )
 
-    delta = 0.2  # TODO: this does not work with 0.1, why? Triage with Salem.
+    delta = 0.1  # TODO: this does not work with 0.1, why? Triage with Salem. ? Salem: Doesn't it ???
```
josephdviviano, line 833: sorry this is an outdated comment and is wrong -- it does work, I had a bug elsewhere when I wrote this.

src/gfn/losses/sub_trajectory_balance.py
```diff
@@ -178,7 +178,7 @@ def get_scores(
             flattening_masks,
         )
 
-    # TODO: inespect whether there exists code which can be factored out in the below
+    # TODO: inspect whether there exists code which can be factored out in the below
```
josephdviviano, line 181: lol oups

josephdviviano (approved):
Looks great! A few minor comments.

saleml:
There are a few tests that don't pass (eg `FAILED testing/test_samplers_and_trajectories.py::test_trajectory_sampling[5-5-0.8-Identity-Box] - ValueError: Sampled actions should be positive`).
I suspect this is because of how the masking and the forward function of `BoxPFNeuralNet` are handled. We can either debug that together.
I will merge anyway. We can either debug that later on, or go with the next PR that uses a very slightly different `BoxPFNeuralNet`

PR #79, opened 2023-07-13
Extending trajectories that are on a gpu device

JosephRRB:
An error occurs when combining two trajectories on gpu using `extend()`. Specifically, when `extend_actions()` tries to pad the `actions` from the trajectories, the padding created is on cpu. So, when `torch.cat()` is called, we get the error:

```python
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
``` 
Presumably, the same will be true for the `log_probs`.

The main change of this PR is putting the paddings for `actions` and `log_probs` on the same device. The test for this change is on `test_extend_trajectories_on_cuda()`.

**Notes:**
- There were some errors in the initial pre-commit involving `flake8` (not found in gitlab) and `isort` (invalid Poetry configuration). I can remove the changes in `pre-commit-config.yaml` in this PR if needed
-  Other file changes are from running `pre-commit`
- The files imported in the tests are located in an external site packages library for my conda environment. So they will not reflect my changes. Instead, I tried to manually import the relevant class in the modified file. Maybe there's a setting during installation so that the imports directly point to the files tracked by git? Do you have other suggestions for this?

Thank you very much!

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
             (?x)^(
                 .*\.ipynb
             )$
-  - repo: https://gitlab.com/PyCQA/flake8
+  - repo: https://github.com/PyCQA/flake8
     rev: 4.0.1
     hooks:
       - id: flake8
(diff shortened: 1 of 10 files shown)


saleml (approved):
Awesome ! Great PR.
Thanks for adding a test for the functionality you wrote.

Note however that in the next few days, v1 will be released, with major API changes. Some of these changes will not be reflected. The tests will for sure.

PR #89, opened 2023-08-01
Refactor modules and parameterizations

josephdviviano:
This PR addresses https://github.com/saleml/torchgfn/issues/77

In particular:

a `GFlowNet` can be instantiated directly, which contains loss calculation and `to_probability_distribution` transformations if required.

Also, `nn.Module` is always subclasses so Pytorch can properly track all `GFlowNet` parameters automatically.

There was furthermore a refactor to increase code reuse. There is no longer a single class for each parameterization -- instead classes are defined according to their general utility. For a concrete example of this see `DiscretePolicyEstimator` which handles forward and backward policy estimators for multiple `GFlowNet`s.


saleml (approved):
LGTM ! Great job

PR #97, opened 2023-08-01
use a separate learning rate for the state flow function for the box environment

saleml:


josephdviviano (approved):
loves it, lgtm, etc.

PR #101, opened 2023-08-02
Device errors

josephdviviano:
+ Added logic to handle moving all tensors to GPU when required. 
+ Does *not* rely on the env to do this 🙃


saleml (approved):
looks great !

Issue #123, opened 2023-08-03
Manually create more involved documentation


josephdviviano:
We need to manually write `.rst` files in the `docs` folder to get a more sophisticated documentation system.

See https://pytorch-forecasting.readthedocs.io/en/stable/ for an example.

vict0rsch:
From #128 -> Transition to relative links instead of absolute links to master

PR #128, opened 2023-08-22
Add generic type annotation for GFlowNet.to_training_samples

marpaia:
Each subclass of `GFlowNet` contains a `to_training_sample` method which may return a different type. This is often a subclass of `Container` like `Trajectories` or `Transitions` or, in the case of `FMGFlowNet`, a `Tuple[State, ...]`. Right now, this is not typed in the `GFlowNet` base class. 

This PR takes a crack at typing this function on the `GFlowNet` base class by adding:

- A new type variable `TrainingSampleType` which is constrained to be either a `Container` subtype or a tuple containing instances of `States`.
- Generic type annotation to the `GFlowNet` class, making it a generic type with respect to the `TrainingSampleType`.
- Corresponding type annotations to the derived classes (`DBGFlowNet`, `ModifiedDBGFlowNet`, `FMGFlowNet`, etc.) to specify the exact type they return from the `to_training_sample` method.
- Test cases in `test_gflownet.py` to verify the correct typing and functionality of the `to_training_samples` method across different `GFlowNet` subclasses.

I came across this because I have a function which returns a `GFlowNet` subclass given a loss function that is defined dynamically via the command line arguments. This looks something like:

```python
def gflownet_for_loss(...) -> GFlowNet:
```

Thus the return type of the function's `to_training_sample` method is undefined. 

While I think that this PR makes the `GFlowNet` base class slightly more clear, there is a downside of this exact implementation that new future implementations of `GFlowNet` subclasses will have to remember to extend the `TrainingSampleType` type, potentially making the process more cumbersome. One could make this slightly more flexible while maintaining some clarity by removing the bounded types on the `TrainingSampleType` type variable.

So all that said, I would understand if the juice isn't worth the proverbial squeeze on this. No pressure, I just figured I'd put this up for consideration.


src/gfn/gflownet/base.py
```diff
@@ -1,18 +1,23 @@
...
 from torchtyping import TensorType as TT
 
 from gfn.containers import Trajectories
+from gfn.containers.base import Container
 from gfn.env import Env
 from gfn.modules import GFNModule
 from gfn.samplers import Sampler
 from gfn.states import States
 
+TrainingSampleType = TypeVar(
```
josephdviviano, line 15: This seems pretty general, I don't see any issues here (@saleml do you?)

src/gfn/gflownet/base.py
```diff
@@ -1,18 +1,23 @@
...
+    "TrainingSampleType", bound=Union[Container, tuple[States, ...]]
+)
 
-class GFlowNet(nn.Module):
+
+class GFlowNet(ABC, nn.Module, Generic[TrainingSampleType]):
```
josephdviviano, line 20: I think this addition requires us to be absolutely sure that all class inheritance is working properly. In many cases we don't specify the `super().__init__()` because `GFlowNet` only inherited from `nn.Module`. In particular we need to ensure that inheritance from `nn.Module` works properly such that things like ` DBGFlowNet.parameters()` still returns all trainable parameters. Did you verify that these changes don't interfere with core pytorch functionality? Perhaps we need more tests to verify this.

testing/test_gflownet.py
```diff
@@ -0,0 +1,64 @@
...
+    gflownet = FMGFlowNet(estimator)
+    mock_trajectories = Trajectories(env)
+    states_tuple = gflownet.to_training_samples(mock_trajectories)
+
+    # Assert that the result is a tuple of `States`
+    assert isinstance(
+        states_tuple, tuple
+    ), f"Expected type tuple, but got {type(states_tuple)}"
+
+    assert isinstance(
```
josephdviviano, line 59: Thanks very much for this PR. I am leaning toward thinking this is a nice addition to the library, and I really appreciate the nice tests (though I'm curious what @saleml thinks as well).

I'm wondering two things:

1) could you please add another test that *asserts that an exception is raised* - I.e., when someone defined the wrong return type in a custom `GFlowNet` class definition, or the GFN returns the wrong type, or something else of this nature. That way we can be sure the library does the right thing when the developer makes a mistake.

2) can we add a short example somewhere (it need not be sensible) that demonstrates clearly what a user needs to do to ensure output types are handled properly?

josephdviviano (commented):
Thanks very much for this PR - it's nice to see the community is interesting in investing in this package! I like where you are going with this - I've left some comments on the PR.

src/gfn/gflownet/base.py
marpaia: Hey @josephdviviano, thanks so much for the feedback and the great point here. I've added a test which checks that both `TBGFlowNet` and `FMGFlowNet` have working methods from `nn.Module` directly accessible. These two classes represent the only two ways that current subclasses of `GFlowNet` come into existence:

- `TBGFlowNet` inherits from `TrajectoryBasedGFlowNet` (and calls `super.__init__`). `TrajectoryBasedGFlowNet` then inherits from `PFBasedGFlowNet` which inherits from `GFlowNet` and calls `super.__init__()`. So thus, while `TrajectoryBasedGFlowNet` doesn't `super.__init__()`, that should be fine because it also doesn't redefine `self.__init__`.
- `FMGFlowNet` inherits from `GFlowNet` directly and calls `super.__init__()`

The test current looks like:

```python
def test_pytorch_inheritance():
    pf_module = BoxPFNeuralNet(
        hidden_dim=32, n_hidden_layers=2, n_components=1, n_components_s0=1
    )
    pb_module = BoxPBNeuralNet(
        hidden_dim=32, n_hidden_layers=2, n_components=1, torso=pf_module.torso
    )

    env = Box()

    pf_estimator = BoxPFEstimator(
        env=env, module=pf_module, n_components=1, n_components_s0=1
    )
    pb_estimator = BoxPBEstimator(env=env, module=pb_module, n_components=1)

    tbgflownet = TBGFlowNet(pf=pf_estimator, pb=pb_estimator)
    assert hasattr(
        tbgflownet.parameters(), "__iter__"
    ), "Expected gflownet to have iterable parameters() method inherited from nn.Module"
    assert hasattr(
        tbgflownet.state_dict(), "__dict__"
    ), "Expected gflownet to have indexable state_dict() method inherited from nn.Module"

    estimator = DiscretePolicyEstimator(env, pf_module, True)
    fmgflownet = FMGFlowNet(estimator)
    assert hasattr(
        fmgflownet.parameters(), "__iter__"
    ), "Expected gflownet to have iterable parameters() method inherited from nn.Module"
    assert hasattr(
        fmgflownet.state_dict(), "__dict__"
    ), "Expected gflownet to have indexable state_dict() method inherited from nn.Module"
```

It would be interesting to consider trying to make this more generic so that it can be more cleanly extended to test other kinds of `GFlowNet` subclasses. In Go, there is a cool pattern called [table tests](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) which would work well here. It's a bit more complicated to do this cleanly here though because each instance of a `GFlowNet` subclass often requires an overlapping set of dependencies so isolating them into their own scope would probably introduce additional code duplication. Happy to iterate though depending on what y'all think is best.

src/gfn/gflownet/base.py
saleml: I like this! I believe the bound can be extended in case there are new losses that come up

src/gfn/gflownet/base.py
```diff
@@ -70,7 +75,7 @@ def sample_trajectories(self, env: Env, n_samples: int) -> Trajectories:
         return trajectories
 
 
-class TrajectoryBasedGFlowNet(PFBasedGFlowNet):
+class TrajectoryBasedGFlowNet(PFBasedGFlowNet[Trajectories]):
```
saleml, line 78: I'm wondering whether there is a way to remove the `TrajectoryBasedGFlowNet` altogether, now that `GFlowNet[Trajectories]` exist. They seem to serve the same purpose.

testing/test_gflownet.py
saleml: This is indeed a nice addition to the library! Thank you for the PR. I agree it would be nice to have a small example somewhere, maybe in the README ?

testing/test_gflownet.py
```diff
@@ -0,0 +1,97 @@
...
+    ), "Expected gflownet to have indexable state_dict() method inherited from nn.Module"
+
+    estimator = DiscretePolicyEstimator(env, pf_module, True)
+    fmgflownet = FMGFlowNet(estimator)
+    assert hasattr(
+        fmgflownet.parameters(), "__iter__"
+    ), "Expected gflownet to have iterable parameters() method inherited from nn.Module"
+    assert hasattr(
+        fmgflownet.state_dict(), "__dict__"
+    ), "Expected gflownet to have indexable state_dict() method inherited from nn.Module"
```
saleml, line 97: I like these tests! Thank you

saleml (commented):
Thanks a lot for this PR @marpaia.

I have left a few comments.

src/gfn/gflownet/base.py
marpaia: I am looking into some way to simplify the class hierarchy and I'm not sure that it is possible. I made this little GraphViz of the class hierarchy of the `gfn.gflownet` module and it looks like `TrajectoryBasedGFlowNet` is necessarily sharing methods like `get_pfs_and_pbs` and `get_trajectories_scores` across `TBGFlowNet`, `SubTBGFlowNet`, and `LogPartitionVarianceGflowNet`:

classes

As you can see, given that every branch node (ie: non-leaf node) in the class tree has greater than one children, I don't think there are any trivial ways to compress the hierarchy without duplicating some functionality of a branch node into its current children.

That said, of course if you have any advice on that or if there is anything specific you'd like to see happen with this class hierarchy in order to simplify it, I am more than happy to take a crack at it! 🙏

testing/test_gflownet.py
marpaia: @saleml are you thinking some documentation to show how to add a new implementation of a `GFlowNet`? If so, I'd be happy to add it to the README, but perhaps it might be better suited to go somewhere like `tutorials/GFLOWNET.md` or something? 

It seems like adding new `GFlowNet` subclasses will not be a super common developer use-case so it might be worth keeping the README to the developer happy paths? Also it seems like it might fit nicely alongside `tutorials/ENV.md` which has similar properties of being a less common, more core development oriented task involving expanding the core object model of the library.

That said, I am of course happy to defer to whatever you would prefer.

README.md
```diff
@@ -50,11 +50,11 @@ pip install .
 
 This repo serves the purpose of fast prototyping [GFlowNet](https://arxiv.org/abs/2111.09266) (GFN) related algorithms. It decouples the environment definition, the sampling process, and the parametrization of the function approximators used to calculate the GFN loss. It aims to accompany researchers and engineers in learning about GFlowNets, and in developing new algorithms.
 
-Currently, the library is shipped with three environments: two discrete environments (Discrete Energy Based Model and Hyper Grid) and a continuous box environment. The library is designed to allow users to define their own environments. See [here](https://github.com/saleml/torchgfn/tree/master/tutorials/ENV.md) for more details.
```
marpaia, line 53: NB: I have updated these links to be relative links instead of hard links to the master branch on GitHub. This has two advantages:

- If on a fork/branch, the links will correctly point to the version of the document on that fork/branch
- If you click the link in a local editor, it will navigate to the local file in the editor instead of via the browser

I think this is somewhat more optimal but I'm happy to revert these changes if it's not prefered.

testing/test_gflownet.py
marpaia: I've taken a crack at adding this in `tutorials/GFLOWNET.md` and linked to it from the main README similarly to `tutorials/ENV.md`. Please feel free to let me know if you'd like more/less/different detail on any aspect of the doc and/or structural changes. 🙏

README.md
```diff
@@ -50,11 +50,11 @@ pip install .
...
+Currently, the library is shipped with three environments: two discrete environments (Discrete Energy Based Model and Hyper Grid) and a continuous box environment. The library is designed to allow users to define their own environments. See [here](./tutorials/ENV.md) for more details.
```
marpaia, line 53: NB: I have updated these links to be relative links instead of hard links to the master branch on GitHub. This has two advantages:

- If on a fork/branch, the links will correctly point to the version of the document on that fork/branch
- If you click the link in a local editor, it will navigate to the local file in the editor instead of via the browser

I think this is somewhat more optimal but I'm happy to revert these changes if it's not prefered!

marpaia: I'm thinking that this might break the docs on https://torchgfn.readthedocs.io/en/latest/ though... Maybe it's worth backing this out and then trying to address this more structurally in #123?

vict0rsch:
(Good to see you here @marpaia 😄)

vict0rsch (approved):

marpaia:
Hey @vict0rsch, it's a pleasure to hack with you again 😄 

It's a really good point that externally defined GFlowNets wouldn't be able to extend the TypeVar if they were returning a different type of training sample. I know we're going off the Python type system deep end on this one but I think the best way to account for this would be to use a [Python Protocol](https://peps.python.org/pep-0544/).

To recap, right now in this PR we have:

```python
TrainingSampleType = TypeVar(
    "TrainingSampleType", bound=Union[Container, tuple[States, ...]]
)
```

Note the (redacted) implementation of `Container`:

```python
class Container(ABC):
    """Base class for states containers (states, transitions, or trajectories)."""

    @abstractmethod
    def __len__(self) -> int:
        ...

    @abstractmethod
    def __getitem__(self, index: int | Sequence[int]) -> Container:
        ...

    ...
```

Basically we have these two types at the moment in our TypeVar: a `Container` and a `Tuple` (of `State`s).

Critically, both types have a length (via `__len__`) and both types are indexable (via `__getitem__`).

So we could add a protocol like:

```python
from typing import Protocol, Any

class TrainingSampleSequence(Protocol):
    def __len__(self) -> int:
        ...

    def __getitem__(self, index: int) -> Any:
        ...

```

And then simply update the type var to:

```python
TrainingSampleType = TypeVar("TrainingSampleType", bound=TrainingSampleSequence)
```

This would ensure that any new `GFlowNet` subclass (inside or outside of the repo) could use ANY indexable type with a length as the training sample.

This is slightly less "safe" because new GFlowNets have more flexibility as to what they can return as training samples but perhaps this is a more accurate representation of what we want the type system to enforce anyway. Plus, the implementation of `GFlowNet` would provide the same exact degree of type system explicitivity by defining the *specific* training sample type.

What do you think? Happy to make this change if you prefer it.

src/gfn/gflownet/base.py
marpaia: Yes, exactly! I think the fact that it wasn't there must have been an oversight so I added it back in there.

README.md
```diff
@@ -150,6 +150,10 @@ open build/html/index.html
 
 See [here](https://github.com/saleml/torchgfn/tree/master/tutorials/ENV.md)
 
+### Defining a GFlowNet
```
saleml, line 153: Thank you for adding GFLOWNET.md. 
Small nitpick, the word GFlowNet can be understood in different ways. Some people would understand "Defining a GFlowNet" as "creating a model then using an existing loss". I believe it makes more sense to add the section this a bit later in the README, and name it some thing like "Extending GFlowNets" or "Creating a new GFlowNet loss", or something like that. Same thing for tutorials/README.md ?

What do you think ?

src/gfn/gflownet/base.py
saleml: Thanks ! Let's keep things as they are then :)

saleml:
> Hey @vict0rsch, it's a pleasure to hack with you again 😄
> 
> It's a really good point that externally defined GFlowNets wouldn't be able to extend the TypeVar if they were returning a different type of training sample. I know we're diving deep into the Python type system on this one but I think the best way to account for this would be to use a [Python Protocol](https://peps.python.org/pep-0544/). Protocols allow for structural subtyping and ensure the presence of certain methods without enforcing inheritance (kind of like an interface in Go).
> 
> To recap, right now in this PR we have:
> 
> ```python
> TrainingSampleType = TypeVar(
>     "TrainingSampleType", bound=Union[Container, tuple[States, ...]]
> )
> ```
> 
> This enforces that TrainingSampleType is or inherits from either `Container` or `Tuple[States, ...]`.
> 
> Note the (redacted) implementation of `Container`:
> 
> ```python
> class Container(ABC):
>     """Base class for states containers (states, transitions, or trajectories)."""
> 
>     @abstractmethod
>     def __len__(self) -> int:
>         ...
> 
>     @abstractmethod
>     def __getitem__(self, index: int | Sequence[int]) -> Container:
>         ...
> ```
> 
> Critically, both `Container` and `Tuple` types have a length (via `__len__`) and both types are indexable (via `__getitem__`).
> 
> So we could add a protocol like:
> 
> ```python
> from typing import Protocol, Any
> 
> class TrainingSampleSequence(Protocol):
>     def __len__(self) -> int:
>         ...
> 
>     def __getitem__(self, index: int | Sequence[int]) -> Any:
>         ...
> ```
> 
> And then simply update the type var to:
> 
> ```python
> TrainingSampleType = TypeVar("TrainingSampleType", bound=TrainingSampleSequence)
> ```
> 
> This would ensure that any new `GFlowNet` subclass (inside or outside of the repo) could use ANY indexable type with a length as the training sample.
> 
> This is slightly less "safe" because new GFlowNets have more flexibility as to what they can return as training samples but perhaps this is a more accurate representation of what we want the type system to enforce anyway. Plus, the implementation of `GFlowNet` would provide the same exact degree of type system explicitivity by defining the _specific_ training sample type.
> 
> What do you think? Happy to make this change if you prefer it.
> 
> EDIT: I implemented this in [c2ee3c3](https://github.com/saleml/torchgfn/commit/c2ee3c338e97d8dbae0f751bab5436c15516120d). I think it is more flexible and a better developer experience but I am happy to revert it if you guys prefer the more explicit definition of types.

I wasn't aware of Protocols, thank you for the explanation !

Not a strong opinion here, but I lean slightly in favour of more explicit typing here, simply because the protocol makes it look like that TrainingSamples need to have a `__len__` and a `__getitem__` methods. In fact, it's merely a coincidence that both `Container`s and `Tuple[States, States]`s have that, and these two methods are not used later on.

If we were to stick with explicit typing, wouldn't we just need to add something to `bound=Union[Container, tuple[States, ...]]` when creating a new GFlowNet?

Thoughts @vict0rsch ?

vict0rsch:
@saleml I get your point. And yes, it's a very simple modification. It's just slightly less friendly when you use a library and you have to modify the source code or ask the devs to create a new release just for you so being flexible by anticipation would make sense too.

That being said, I'm in favour of not spending too much time solving issues that have not happened yet 😅 So if you have a preference for types, that's fine with me!

marpaia:
OK, thanks for all of the feedback guys, it's really kind and very much appreciated!

I've made two changes:

- I backed out the changes where the TypeVar bounds to a Protocol. It's back to bounding explicit types via the Union.
- I moved the docs that I added to the end of the README under a section "Extending GFlowNets".

@saleml I would understand if you're motivated to keep the README docs more succinct so please don't hesitate to let me know if you'd prefer that section to be more concise and/or if you want me to add the full content back to `GFLOWNET.md` with a more minimal version in the README, etc.

saleml (approved):
LGTM! Thank you for the changes.

marpaia:
Awesome, thanks for the super thorough review and fun discussion. The PR is only mergeable by people with write access to the repo so I'll leave it to you guys to merge whenever works for you all! 🙏

josephdviviano:
Thanks for the great work!

josephdviviano:
Sorry to say but I think we need to revisit this work - @marpaia - I believe it introduced a bug that breaks all of our examples (https://github.com/saleml/torchgfn/tree/master/tutorials/examples) 

please see for an example https://github.com/saleml/torchgfn/issues/137

Issue #134, opened 2023-08-30
`DiscretePolicyEstimator` has no printable representation


brunorousseau:
The class `DiscretePolicyEstimator` derives from `GFNModule`. This parent class implements `__repr__` by making explicit reference to an environment: 
![Screenshot 2023-08-30 at 3 19 57 PM](https://github.com/saleml/torchgfn/assets/106997596/122fc3b8-a3f1-4b1c-aa44-96a592815ecd)

The child class inherits this function, but has no `env` property. Thus, instances of the child class cannot be printed:
![Screenshot 2023-08-30 at 3 18 54 PM](https://github.com/saleml/torchgfn/assets/106997596/fc38f4e0-6768-4ab9-8297-4f5b23d7b308)

saleml:
Thank you for raising the issue. It will be fixed in an imminent release.

PR #135, opened 2023-09-01
fix __repr__ of modules 

saleml:
This PR fixes the issue raised in #134. 

Other ideas of a `__repr__` ?


josephdviviano (approved):
can think of a more complete representation in the future (all nn.Module stuff plus the added base features of a GFNModule) but for now this is great

PR #139, opened 2023-10-04
Fix tests

josephdviviano:
This adds `Generic` to `PFBasedGFlowNet` and `TrajectoryBasedGFlowNet`, previously these imports were broken.

Note I am not 100% sure that `TrainingSampleType` is correct for both of these classes, I don't fully understand how to use `Generic` properly.

### DIFF
--- a/src/gfn/gflownet/base.py
+++ b/src/gfn/gflownet/base.py
@@ -55,7 +55,7 @@ class GFlowNet(ABC, nn.Module, Generic[TrainingSampleType]):
         """Computes the loss given the training objects."""
 
 
-class PFBasedGFlowNet(GFlowNet):
+class PFBasedGFlowNet(GFlowNet, Generic[TrainingSampleType]):
     r"""Base class for gflownets that explicitly uses $P_F$.
 
     Attributes:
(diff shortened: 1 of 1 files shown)


saleml (approved):
Thanks for the fix. All the tests pass now.

marpaia:
Hey guys, I'm sorry for the delay in responding here, it's been a busy week at work with a lot of GitHub notifications 😅 

I don't think this PR is quite right. `PFBasedGFlowNet` shouldn't need to redefine the `Generic[TrainingSampleType]` as a subclass because it is already exactly subclassed by `GFlowNet`. Further, I think that `TrajectoryBasedGFlowNet(PFBasedGFlowNet[Trajectories])` is the right signature for `TrajectoryBasedGFlowNet` because it is a subclass of `PFBasedGFlowNet` which uses `Trajectories` as the type returned from `to_training_samples`.

I've tried to revert this commit locally and reproduce the error but I cannot. Are there more concrete reproduction steps for that? I'd be happy to take a crack at it.

josephdviviano:
Thanks!

No other steps. If you are on master and using the exact environment in the
pyproject.toml, the tests will fail. This was also reproduced by some of
our collaborators at Intel.

Could you send your environment? We can diff with mine to see if there's
anything there. It could be the python version behaviour changing slightly.

Thanks for offering your help. I don't really understand this typing scheme
yet.

Let me see if I can get you into the Mila slack as a guest!
Joseph Viviano
viviano.ca

josephdviviano:
@marpaia can you send me your email so I can add you to the mila slack

marpaia:
Awesome, thanks @josephdviviano, I've just sent you an email!

Issue #155, opened 2024-02-13
Nit: `log_reward_clip_min` isn't defined in `gflownet/base.py` - it depends on definitions in the subclasses.


josephdviviano:
in `base.py`

```
        log_rewards = trajectories.log_rewards
        # TODO: log_reward_clip_min isn't defined in base ().
        if math.isfinite(self.log_reward_clip_min) and log_rewards is not None:
            log_rewards = log_rewards.clamp_min(self.log_reward_clip_min)

```

this logic corresponds to an attribute only defined in subclasses. I don't think this is best practice, and the corresponding subclass logic should be centralized in `base.py`.

josephdviviano:
PR created, closing.

PR #158, opened 2024-02-16
nit: added default value for log_reward_clip_min in abstract base class

josephdviviano:
Addresses https://github.com/GFNOrg/torchgfn/issues/155, we now have a default value for `log_reward_clip_min` defined in the abstract base class.


saleml (approved):
ok ok

PR #159, opened 2024-02-17
device handling fix

josephdviviano:
Removes a "feature" whereby the `NeuralNet` would automatically cast tensors to a particular device. This would have the strange side effect of moving a part of the `gflownet` instance to an unexpected device, causing a bug that is confusing to track down.

Instead, the code will now fail in a sensible way if the user does not cast tensors to the correct device, making debugging easy.

This might require slightly more thought in the case where one has multiple environments running on difference devices, but we can address this in a follow up PR.


saleml:
Essentially, with this change, we would get the usual pytorch error of incompatible devices instead, right?

josephdviviano:
Yea exactly, it’s up to the user to ensure everything is on the correct
device explicitly

Joseph (Mobile)

saleml (approved):

Issue #136, opened 2023-09-20
[help wanted] Traning a DiscreteEBM ends up with "Log probabilities are inf. This should not happen." 


ermiaetemadi:
Hi. I'm trying to train a DiscreteEBM environment for square lattice ising model. It's working fine with small grid lengths but when I attempt to increase the grid length i get this:
`RuntimeError: Log probabilities are inf. This should not happen.`

I can search for the appropriate batch size for each length with trial and error but i can't understand why it is happening. 

My code:  

```python
import torch

from tqdm import tqdm
import wandb
from argparse import ArgumentParser


from gfn.gym import DiscreteEBM
from gfn.gym.discrete_ebm import IsingModel
from gfn.gflownet import FMGFlowNet
from gfn.utils.modules import NeuralNet
from gfn.modules import DiscretePolicyEstimator
from gfn.utils.common import validate

def main(args):
    
    # Configs

    use_wandb = len(args.wandb_project) > 0
    if use_wandb:
        wandb.init(project=args.wandb_project)
        wandb.config.update(args)

    device =  "cpu"
    torch.set_num_threads(args.n_threads)

    hidden_dim = 512
    n_hidden = 2
    acc_fn = "relu"
    lr = 0.001
    lr_Z = 0.01
    L = args.L

    validation_samples = 1000

    # Ising model parameters

    def ising_n_to_ij(L, n):

        i = n // L
        j = n - i*L

        return (i, j)


    N = L**2
    J = torch.zeros((N, N), device=torch.device(device))
    for k in range(N):
        for m in range(k):
        
            x1, y1 = ising_n_to_ij(L, k)
            x2, y2 = ising_n_to_ij(L, m) 
            if x1 == x2 and abs(y2 - y1) == 1:
                J[k][m] = 1
                J[m][k] = 1
            elif y1 == y2 and abs(x2 - x1) == 1:
                J[k][m] = 1
                J[m][k] = 1
                
    for k in range(L):
     
        J[k*L][(k+1)*L - 1] = 1
        J[(k+1)*L - 1][k*L] = 1
        J[k][k+N-L] = 1
        J[k+N-L][k] = 1

    J = args.J * J

    # Ising model env

    ising_energy = IsingModel(J)
    ising_env = DiscreteEBM(N, alpha=1, energy=ising_energy, device_str=device)

    # Parametrization and losses

    pf_module = NeuralNet(
                    input_dim=ising_env.preprocessor.output_dim,
[...]

PR #162, opened 2024-02-20
added ising example

josephdviviano:
Added a simple example of training an ising model using the current API and a flow matching model (from https://github.com/GFNOrg/torchgfn/issues/136).


saleml:
How long does it take you to run the script with the default parameters? (L=16)

josephdviviano:
Many hours - we could turn that down, it’s true

Joseph (Mobile)

saleml (approved):

PR #163, opened 2024-02-22
Dont recompute masks

josephdviviano:
This isn't working @saleml  -- please see the issue in `samplers.py`. 

You can reproduce the error with `tutorials/examples/train_hypergrid_simple.py`

### DIFF
--- a/src/gfn/states.py
+++ b/src/gfn/states.py
@@ -417,7 +417,6 @@ class DiscreteStates(States, ABC):
             exit_idx = torch.ones(self.batch_shape + (1,)).to(cond.device)
         self.forward_masks[torch.cat([cond, exit_idx], dim=-1).bool()] = False
 
-
     def set_exit_masks(self, batch_idx):
         """Sets forward masks such that the only allowable next action is to exit.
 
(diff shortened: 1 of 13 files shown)


saleml:
Some debugging:

First, I changed the batch_size to 3 in the script. Then, with a breakpoint at the assertion error, I see that the forward masks of all the steps within the batch of 3 trajectories are the same. 

So here is the interesting part. If I add a breakpoint at lines 195 and 199 of `samplers.py`, I get the following:

Screenshot 2024-02-24 at 12 58 53 PM

The first mask is ok. Initially, we have 3 copies of s0, so the masks should all be True. Once we call `_step`, we haven't explicitly modified the list `trajectories_states_b`. Yet, the forward_masks of its only element changed, just because we called `_step`.

So the problem happens here in `env.py`
Screenshot 2024-02-24 at 1 04 10 PM


And looking at the `update_masks` function implemented in Hypergrid:

Screenshot 2024-02-24 at 1 04 53 PM

it seems to me that the masks are changed in place and that the problem is due to https://github.com/GFNOrg/torchgfn/pull/149. I don't remember if tests were passing in that PR (I haven't reviewed that PR).

Let me know what you think

src/gfn/states.py
```diff
@@ -446,3 +446,20 @@ def init_forward_masks(self, set_ones: bool = True):
...
+def stack_states(states: List[States]):
+    """Given a list of states, stacks them along a new dimension (0)."""
+    state_example = states[0]  # We assume all elems of `states` are the same.
+
+    stacked_states = state_example.from_batch_shape((0, 0))  # Empty.
+    stacked_states.tensor = torch.stack([s.tensor for s in states], dim=0)
+    if state_example._log_rewards:
+        stacked_states._log_rewards = torch.stack([s._log_rewards for s in states], dim=0)
+    stacked_states.forward_masks = torch.stack([s.forward_masks for s in states], dim=0)
+    stacked_states.backward_masks = torch.stack([s.backward_masks for s in states], dim=0)
```
saleml, line 460: this should only be implemented for DiscreteStates, not all States

josephdviviano:
Good catch! These tests pass fine - I think the inplace update of masks is desirable behaviour except in this case where we want to accumulate a trajectory of states.

To reduce user error, the base `States` class could have a `clone` method, which `return deepcopy(self)`.

When I was messing around, I tried `copying` the states. I should have instead used `deepcopy`, which prevents the forward masks from being updated inplace.

josephdviviano:
OK @saleml figured it out - check line 413 here https://github.com/GFNOrg/torchgfn/pull/163/commits/77e7e1b524a0a640f51b30a82a73a5ea8fee9e90

Before setting the `False` elements, since we are doing inplace operations, we must first set all values to `True` to prevent side effects over multiple steps (`self.forward_masks[:] = True`).

This, plus using `deepcopy` where appropriate, fixes the issue, and we no longer recompute masks.

saleml (approved):

Issue #168, opened 2024-02-27
Replay buffer broken in `train_hypergrid.py`


josephdviviano:
When a replay buffer is being used, the gflownet will not converge, but will if there is no replay buffer. I suspect this has something to do with off-policy logic changes that were introduced here:

https://github.com/GFNOrg/torchgfn/pull/147

Issue #156, opened 2024-02-13
`Transitions` container should then have some `estimator_outputs` attribute to avoid duplicate computation.


josephdviviano:
In `detailed_balance.py`, we have:

```
        if not self.off_policy:
            valid_log_pf_actions = transitions.log_probs
        else:
            # Evaluate the log PF of the actions sampled off policy.
            # I suppose the Transitions container should then have some
            # estimator_outputs attribute as well, to avoid duplication here ?
            module_output = self.pf(states)  # TODO: Inefficient duplication.
            valid_log_pf_actions = self.pf.to_probability_distribution(
                states, module_output
            ).log_prob(
                actions.tensor
            )  # Actions sampled off policy.
```

We could aboid this second forward pass of `.pf()` by storing the estimator outputs in the transitions class.

Ideally, both `Trajectories` and `Transitions` would be able to access the same estimator outputs in memory if there were ever a need to keep track of both.

PR #174, opened 2024-03-21
Fix off policy

saleml:
This fixes https://github.com/GFNOrg/torchgfn/issues/168.
The idea is to remove the arguments we had before `off_policy` and `sample_off_policy`, and be explicit about what we're evaluating and storing when sampling.
When being on_policy, we should store the logprobs. This is the default.
When being off_policy, with a tempered/modified PF, we should only store `estimator_outputs`.
When we use a replay buffer, we don't need to store anything - we should recalculate the logprobs.

Additionally, this fixes FM + ReplayBuffer, that was broken before, because states extension didn't take into account the `_log_probs` attribute.


src/gfn/gflownet/detailed_balance.py
```diff
@@ -66,19 +72,20 @@ def get_scores(self, env: Env, transitions: Transitions) -> Tuple[
 
         if states.batch_shape != tuple(actions.batch_shape):
             raise ValueError("Something wrong happening with log_pf evaluations")
-        if not self.off_policy:
+        if (
+            transitions.log_probs is not None
```
josephdviviano: I'm seeing this logic a few times in the code. Should we abstract it into a utility like 
```
def has_log_probs(obj):
    return obj.log_probs is not None and obj.log_probs.nelement() > 0
```
?

josephdviviano: I've added this utility function.

src/gfn/gflownet/detailed_balance.py
```diff
@@ -66,19 +72,20 @@ def get_scores(self, env: Env, transitions: Transitions) -> Tuple[
...
+            and transitions.log_probs.nelement() > 0
+            and not recalculate_all
+        ):
             valid_log_pf_actions = transitions.log_probs
         else:
-            # Evaluate the log PF of the actions sampled off policy.
-            # I suppose the Transitions container should then have some
-            # estimator_outputs attribute as well, to avoid duplication here ?
-            # See (#156).
```
josephdviviano, line 75: Why did you remove this issue reference (#156) ?

src/gfn/gflownet/trajectory_balance.py
```diff
@@ -53,7 +53,9 @@ def loss(self, env: Env, trajectories: Trajectories) -> TT[0, float]:
             ValueError: if the loss is NaN.
         """
         del env  # unused
-        _, _, scores = self.get_trajectories_scores(trajectories)
+        _, _, scores = self.get_trajectories_scores(
+            trajectories, recalculate_all=recalculate_all
```
josephdviviano, line 57: I'm wondering if there's a more explicit name for `recalculate_all` -- like `recalculate_all_logprobs`?

src/gfn/samplers.py
```diff
@@ -89,38 +93,37 @@ def sample_actions(
...
+                is useful for off-policy training with tempered policy.
+            save_logprobs: If True, calculates and saves the log probabilities of sampled
+                actions. This is useful for on-policy training.
             policy_kwargs: keyword arguments to be passed to the
                 `to_probability_distribution` method of the estimator. For example, for
                 DiscretePolicyEstimators, the kwargs can contain the `temperature`
                 parameter, `epsilon`, and `sf_bias`. In the continuous case these
                 kwargs will be user defined. This can be used to, for example, sample
                 off-policy.
-            debug_mode: if True, everything gets calculated.
```
josephdviviano, line 114: Why is `debug_mode` removed? If I recall, this was important for tests.

josephdviviano (commented):
I really like these API changes -- I have a few small questions before we approve (but this might not require any further changes to the code -- I just want to understand).

src/gfn/samplers.py
saleml: Isn't this the same as `recalculate_all`?

src/gfn/utils/common.py
saleml, line 1: Good idea! Can the function be a class method of `Container`? 
`self.has_log_prob()` looks more natural than `has_log_prob(self)`

src/gfn/env.py
```diff
@@ -393,7 +393,7 @@ class DiscreteEnvStates(DiscreteStates):
 
     def make_actions_class(self) -> type[Actions]:
         env = self
-        n_actions = self.n_actions
+        self.n_actions
```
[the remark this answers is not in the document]
saleml: Not sure what happened. Actually, we don't need that line altogether (thanks Pylance) !
I'm removing the whole line

src/gfn/gflownet/detailed_balance.py
saleml: My bad! Added back

src/gfn/gflownet/trajectory_balance.py
saleml: yes, good idea, done

src/gfn/samplers.py
josephdviviano: right -- I'll change it back and add a note :)

src/gfn/utils/common.py
josephdviviano: Hmm, the only issue is we actually use it in `TrajectoryBasedGFlowNet`

josephdviviano (approved):
lgtm!

PR #175, opened 2024-03-30
Prioritized replay buffer

josephdviviano:
I've adde a prioritized replay buffer. This:

- Only adds examples if their reward is larger than the min reward found in the buffer.
- Only adds examples if they are unique (by default).

In general, uniqueness is defined as distances between the candidate batch and buffer states using a `p_norm` -- this is configurable by the user. The default settings use a p-norm of 1 and a distance threshold of 0, i.e., the added states should not be identical to any state already in the buffer.

**Important** you can test this using 

```
tutorials/examples/train_hypergrid.py --replay_buffer_size 1000 --replay_buffer_prioritized
```

Note: currently, the standard buffer outperforms the prioritized buffer using these default settings!

+ Standard: `'loss': 9.107343066716567e-05, 'states_visited': 998416, 'l1_dist': 0.00023296871222555637, 'logZ_diff': 0.001130819320678711`
+ Prioritized: `'loss': 0.0003514138516038656, 'states_visited': 998416, 'l1_dist': 0.00017267849761992693, 'logZ_diff': 0.0020639896392822266`

In the debugger, I could determine that **no samples were ever added to the buffer after it was originally filled, because the states were not found to be unique**. I.e., in `replay_buffer.py` the following logic always had `idx_batch_buffer` as completely full of `False`:

```
    # Remove non-diverse examples according to the above distances.
    idx_batch_batch = batch_batch_dist > self.cutoff_distance
    idx_batch_buffer = batch_buffer_dist > self.cutoff_distance
    idx_diverse = idx_batch_batch & idx_batch_buffer
```

@saleml I'd be curious to get your opinion on this. Perhaps we can tweak the implementation of the prioritized replay buffer, or perhaps this should be expected behaviour for this relatively simple example. I am not sure.

### DIFF
--- a/src/gfn/containers/replay_buffer.py
+++ b/src/gfn/containers/replay_buffer.py
@@ -18,7 +18,6 @@ class ReplayBuffer:
 
     Attributes:
         env: the Environment instance.
-        loss_fn: the Loss instance
         capacity: the size of the buffer.
         training_objects: the buffer of objects used for training.
         terminating_states: a States class representation of $s_f$.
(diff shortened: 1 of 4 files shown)


src/gfn/containers/trajectories.py
```diff
@@ -232,22 +232,34 @@ def extend(self, other: Trajectories) -> None:
...
 
+        # Concatenate log_rewards of the trajectories.
         if self._log_rewards is not None and other._log_rewards is not None:
             self._log_rewards = torch.cat(
                 (self._log_rewards, other._log_rewards),
                 dim=0,
             )
+        # If the trajectories object does not yet have `log_rewards` assigned but the
+        # external trajectory has log_rewards, simply assign them over.
+        elif self._log_rewards is None and other._log_rewards is not None:
```
saleml, line 251: why is this needed?
This is actually dangerous and can easily lead to undesired behavior.

src/gfn/containers/trajectories.py
```diff
@@ -232,22 +232,34 @@ def extend(self, other: Trajectories) -> None:
...
         else:
             self._log_rewards = None
 
+        # Ensure log_probs/rewards are the correct dimensions. TODO: Remove?
+        if self.log_probs.numel() > 0:
+            assert self.log_probs.shape == self.actions.batch_shape
+
+        if self.log_rewards is not None:
+            assert len(self.log_rewards) == self.actions.batch_shape[-1]
+
```
saleml, line 262: fair

src/gfn/containers/replay_buffer.py
```diff
@@ -105,3 +104,137 @@ def load(self, directory: str):
...
+        """Adds a training object to the buffer."""
+        terminating_states = None
+        if isinstance(training_objects, tuple):
+            assert self.objects_type == "states" and self.terminating_states is not None
+            training_objects, terminating_states = training_objects
+
+        to_add = len(training_objects)
+
+        self._is_full |= self._index + to_add >= self.capacity
+        self._index = (self._index + to_add) % self.capacity
```
saleml, line 175: But if we don't add all the training, we would have increased self._index by more than needed.

Acutally, do we need `self._index` at all in ReplayBuffers?

src/gfn/containers/trajectories.py
josephdviviano: We had a situation in the buffer where the empty initialized trajectory had `_log_rewards = None`, so any call to `.extend()` did not update the `_log_rewards` - we can handle this a few ways but, as is, thankfully tests are passing.

src/gfn/containers/trajectories.py
josephdviviano: yes good to check -- but maybe ideal that we don't have to check -- this adds overhead.

src/gfn/containers/replay_buffer.py
josephdviviano: It isn't clear to me what this is used for actually. We can chat about it on our meeting.

josephdviviano:
@saleml Just need your official sign off on this before I can merge :)

saleml (approved):
LGTM! Great PR

Issue #173, opened 2024-03-21
Is it normal that running `train_line.py` renders samples from one mode only?


saleml:

josephdviviano:
I likely need to change the default options to enable off policy exploration

josephdviviano:
![Screenshot 2024-03-31 at 12 33 33](https://github.com/GFNOrg/torchgfn/assets/4142570/be2b3efb-19ce-4cde-91fb-a8e76e97c391)

Hmm, on my machine, the training isn't perfect (the default options undertrain the policy) but I definitely sample from both modes.

Were your experiments off of `master` or is this maybe related to changes we made RE: off policy training in the other open pr (https://github.com/GFNOrg/torchgfn/pull/174)?

saleml:
I'm investigating the issue. Indeed, different behavior on `mater` and on `fix_off_policy`.

One thing worth noting is that with the original number of trajectories (1.28e6), sure the samples are from one mode only, but are more accurate.


![image](https://github.com/GFNOrg/torchgfn/assets/16039448/94dfb6a2-9635-4f57-a768-dfddad95b4bd)

TBC

saleml:
In `fix_off_policy`, I had forgotten a keyword. The problem is fixed in https://github.com/GFNOrg/torchgfn/pull/174/commits/89c72b5add431fcd8a787323cdc14aee7ee1ffe8

PR #176, opened 2024-03-31
Train line bugfix

josephdviviano:
(maybe) addresses https://github.com/GFNOrg/torchgfn/issues/173


saleml:
#173 is solved (via #174). I reverted back to the original number of trajectories. Kept your modification for matplotlib.
Ready to merge if you're ok with it !

saleml (approved):

Issue #153, opened 2024-02-08
About having graph in the States


wang422003:
Hello and thank you for developing the torchgfn package! It's been a fantastic resource, and I appreciate all the work that's gone into it.

I'm exploring the package and am particularly interested in integrating graph data within the states, specifically including node features and the adjacency matrix. However, I'm currently unsure of the best approach to achieve this within the framework of torchgfn.

Could you provide any guidance, instructions, or suggestions on how to incorporate graph structures into the states? Any help or pointers towards relevant documentation or examples would be greatly appreciated!

josephdviviano:
This is something we're also interested in building into torchgfn.

I believe we would need to take inspiration from https://github.com/recursionpharma/gflownet to do it properly. 

This is one of a few major features that need to be added to the package for it to be fully featured. Would you be interested in assisting on the development of this?

wang422003:
Many thanks for the reply! I will definitely check the link. 

I am really interested in GFlowNet and interested in assisting on the development of getting graphs in the states!

josephdviviano:
@saleml for reference : https://github.com/recursionpharma/gflownet/blob/ef5f2cbb14f3f35bc2b95d33b9b3cf86cc558d17/src/gflownet/envs/graph_building_env.py#L18

@wang422003 for now, if you want a library that works well with graphs, you can check the above work from Emmanuel Bengio's group at Recursion. We plan to support similar functionality but it will take us some time to decide exactly how.

wang422003:
> @wang422003 for now, if you want a library that works well with graphs, you can check the above work from Emmanuel Bengio's group at Recursion. We plan to support similar functionality but it will take us some time to decide exactly how.

Many thanks for sharing. Actually, I checked the link and figured out that their current implementation only supports undirected graphs. So I am now thinking about how to make it work for directed graphs as well.

josephdviviano:
We would love to collaborate with you on this, this would be quite useful for multiple domains and some older work that was never successfully integrated e.g., https://github.com/GFNOrg/torchgfn/pull/26

saleml:
addressed in https://github.com/GFNOrg/torchgfn/pull/210

josephdviviano:
@wang422003 please follow the above issue for updates.

PR #183, opened 2024-08-23
add support for including Graphs as States of GFlowNet

ashdtu:
## Describe your changes
Type: **Feature**
Add support for including Graphs as States of GFlowNet
The Graph structure is represented via the `Data` class of [Torch Geometric](https://pytorch-geometric.readthedocs.io/en/latest/modules/data.html). The GraphStates object is represented via the `Batch` class which encapsulates batching of `Data` objects and their efficient indexing. As opposed to the existing States object which requires appending dummy states for batching different length Trajectories together, we seek to support Trajectories by representing it as nested `Batch` object. 

The existing implementation of Trajectory supports the following indexing dimensions: `(Num time steps, Num trajectories, State Size)`.  The nested _Batch of Batch_ object representing state Trajectories would naturally indexing of the form (Num trajectories, Num timesteps, State size), and would need to implement logic for flipping the indexing dimensions internally in `_getitem_()` and `_setitem_()`. 

## Issue Number
#153 

## Testing
- [ ] Unit tests for helper functions of GraphState class: initialization, appending,  batching etc. 
- [ ] Compatibility check with Trajectories, Transition class 
- [ ] Unit testing with FlowMatching : accessing GraphStates directly in loss function
- [ ] Unit testing with TrajectoryBalance: accessing GraphStates via trajectory class

### DIFF
--- a/src/gfn/states.py
+++ b/src/gfn/states.py
@@ -3,10 +3,11 @@ from __future__ import annotations  # This allows to use the class name in type 
 from abc import ABC
 from copy import deepcopy
 from math import prod
-from typing import Callable, ClassVar, List, Optional, Sequence, cast
+from typing import Callable, ClassVar, List, Optional, Sequence, cast, Tuple
 
 import torch
 from torchtyping import TensorType as TT
+from torch_geometric.data import Batch, Data
 
 
 class States(ABC):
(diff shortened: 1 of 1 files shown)


LicoriceLin:
Hi, is there any further plan on implementing this feature? What's the bugs so far? Is there anything I could help out?

And if I just want to hack it, can I simply copy the `GraphStates` class as a temporary solution?

josephdviviano:
Hi @LicoriceLin yes there is a plan and it's now a top priority. A few of us will be looking at adding this feature starting next week and we would love your help and or feedback on the implementation to ensure we're able to support your needs.

PR #188, opened 2024-09-25
conditional gfn

josephdviviano:
Supports conditioning on a tensor of `shape=[n_trajectories, n_cond_dims]`. This is passed by the user during a call to the sampler.

Implemented for all GFlowNets. Note that the current version expects a particular kind of estimator. I can imagine this will lead to future changes - e.g., we should have some `Estimators` which expect huggingface models, so we can use them to produce conditioning vectors / to initialize the policy (this will obviously be a future PR).

Note that the conditioning is useless in my example, we should have a better use-case envisioned for the demo. The demo currently is not complete for all GFlowNet types.


josephdviviano:
Don't worry about the tests - they should be easy to fix.

I can make the chances for DB, Sub-TB, and FM pretty easily if we agree this is a good approach, before a proper review.

src/gfn/modules.py
```diff
@@ -193,6 +199,56 @@ def to_probability_distribution(
...
+
+
+class ConditionalDiscretePolicyEstimator(DiscretePolicyEstimator):
+    r"""Container for forward and backward policy estimators for discrete environments.
+
+    $s \mapsto (P_F(s' \mid s, c))_{s' \in Children(s)}$.
+
+    or
+
+    $s \mapsto (P_B(s' \mid s, c))_{s' \in Parents(s)}$.
```
saleml, line 214: might be worth mentioning that this is a s very specific conditioning use-case, where the condition is encoded separately, and embeddings are concatenated.

I don't think we can do a generic one, but this should be enough as an example !

src/gfn/samplers.py
```diff
@@ -68,7 +67,28 @@ def sample_actions(
                 the sampled actions under the probability distribution of the given
                 states.
         """
-        estimator_output = self.estimator(states)
+        # TODO: Should estimators instead ignore None for the conditioning vector?
```
saleml, line 70: wouldn't it be cleaner with fewer if else blocks ?

src/gfn/modules.py
josephdviviano: What other conditioning approaches would be worth including? Cross attention?

josephdviviano: In general I would think the conditioning should be embedded / encoded separately --- or would the conditioning just need to be concatenated to the state before input? I could add support for that.

saleml: I don't think there is an exhaustive list of ways we can process the condition. What you have is great as an example. I suggest you just add a comment or doc that the user might want to write their own module

src/gfn/samplers.py
saleml: Ok ! makes sense.

saleml:
LGTM! Looking forward to test this feature

src/gfn/samplers.py
josephdviviano: I added these `exception_handlers` to reduce the cruft.

src/gfn/gflownet/base.py
```diff
@@ -32,41 +35,41 @@ class GFlowNet(ABC, nn.Module, Generic[TrainingSampleType]):
     def sample_trajectories(
         self,
         env: Env,
-        n_samples: int,
```
saleml, line 35: it looks like you're handling the conditioning input to this function as a kwarg, whereas `sampler`'s `sample_trajectories` have an explicit `conditioning` input. I'm wondering if you have a particular reason for this choice

tutorials/examples/train_conditional.py
```diff
@@ -0,0 +1,236 @@
...
+
+    n_iterations = int(10)  #1e4)
+    batch_size = int(1e4)
+
+    print("+ Training Conditional {}!".format(type(gflownet)))
+    for i in (pbar := tqdm(range(n_iterations))):
+        conditioning = torch.rand((batch_size, 1))
+        conditioning = (conditioning > 0.5).to(torch.float)  # Randomly 1 and zero.
+
+        trajectories = gflownet.sample_trajectories(
```
saleml, line 194: Screenshot 2024-10-05 at 4 36 56 PM

pylance is not happy with these two variables. I'm wondering if this is due to using **kwargs (see my comment in the `base.py` file). If so, it would be nice to decide whether we should not care hereafter about pylance and co, or care, in which case, no more kwargs in the code base

tutorials/examples/train_conditional.py
```diff
@@ -0,0 +1,236 @@
...
+    else:
+        print("What is this gflownet? {}".format(type(gflownet)))
+
+    n_iterations = int(10)  #1e4)
+    batch_size = int(1e4)
+
+    print("+ Training Conditional {}!".format(type(gflownet)))
+    for i in (pbar := tqdm(range(n_iterations))):
+        conditioning = torch.rand((batch_size, 1))
+        conditioning = (conditioning > 0.5).to(torch.float)  # Randomly 1 and zero.
```
saleml, line 192: IIUC, the conditioning doesn't change anything in this example. While this file is a great way to show how one can code their conditional gflownet, what do you think of slightly altering the setting here, e.g. by making the environment conditional (e.g. hide one of the 4 modes if conditioning=1), and then, post-training, have some validation where we compare the resulting pair of distributions to the pair of target distributions ?

tutorials/examples/train_conditional.py
```diff
@@ -0,0 +1,236 @@
...
+    train(environment, gflownet)
+
+    gflownet = build_db_mod_gflownet(environment)
+    train(environment, gflownet)
+
+    gflownet = build_subTB_gflownet(environment)
+    train(environment, gflownet)
+
+    gflownet = build_fm_gflownet(environment)
+    train(environment, gflownet)
```
saleml, line 232: argparse this?

src/gfn/gflownet/trajectory_balance.py
saleml, line 1: I'm pleasantly surprised no change is needed for the LogPartitionVarianceLoss. Right?

saleml (commented):
I'm happy to see this being added to the library. Great work! Great code design, and thanks for factorizing a few other things, including the context managers / error handlers.

I left a few comments and a suggestion for the script.

src/gfn/gflownet/base.py
josephdviviano: I think maybe all functions should use an explicit `conditioning` kwarg, what do you think?

saleml: I agree that it would be cleaner

josephdviviano: it should be done now

tutorials/examples/train_conditional.py
josephdviviano: so I think I fixed this but moving to `**kwargs: Any` but we have a multitude of other harder to handle `pylance` issues that I'm not sure what to do about and warrants a discussion bigger than the scope of this PR, I think.

tutorials/examples/train_conditional.py
josephdviviano: yes, you're right. can we save this for a follow up PR?

josephdviviano: I filed the issue here. if you agree I'd like to do this work separately. 

https://github.com/GFNOrg/torchgfn/issues/190

tutorials/examples/train_conditional.py
josephdviviano: fini

src/gfn/gflownet/trajectory_balance.py
josephdviviano: I don't think we need the conditioning information here, and I agree it's nice that the code naturally reflected that. Please correct me if I misunderstand this loss.

src/gfn/gflownet/flow_matching.py
```diff
@@ -198,8 +205,11 @@ def loss(
         )
         return fm_loss + self.alpha * rm_loss
 
-    def to_training_samples(
-        self, trajectories: Trajectories
-    ) -> tuple[DiscreteStates, DiscreteStates, torch.Tensor]:
+    def to_training_samples(self, trajectories: Trajectories) -> Union[
+        Tuple[DiscreteStates, DiscreteStates, torch.Tensor, torch.Tensor],
+        Tuple[DiscreteStates, DiscreteStates, None, None],
```
saleml, line 210: 🤯

src/gfn/modules.py
```diff
@@ -240,13 +240,20 @@ def __init__(
         self.conditioning_module = conditioning_module
         self.final_module = final_module
 
-    def forward(
-        self, states: States, conditioning: torch.tensor
+    def _forward_trunk(
```
saleml, line 243: is what you call `trunk` the same thing I called `torso` before ?

saleml:
LGTM!

Thanks for the PR

saleml (approved):

src/gfn/modules.py
josephdviviano: Yeah -- let me unify the naming

PR #202, opened 2024-10-18
Fix add for PrioritizedReplayBuffer

younik:
Fix for replay buffer to also adds terminating states (when PrioritizedReplayBuffer handles tuple of states)


saleml (approved):
Thanks for the fix. LGTM

PR #203, opened 2024-10-18
Minor refactorings and Rename NeuralNet to MLP

hyeok9855:
This commit contains:
- several incremental refactorings
- remaning `NeuralNet` to `MLP`, considering the broader concept of neural network.


josephdviviano:
I think you just need to resolve conflict with recent updates to `master`

src/gfn/containers/replay_buffer.py
```diff
@@ -40,17 +40,15 @@ def __init__(
         self.env = env
         self.capacity = capacity
         self.terminating_states = None
+        self.objects_type = objects_type
```
josephdviviano, line 43: good catch

src/gfn/env.py
```diff
@@ -275,6 +275,13 @@ def log_partition(self) -> float:
             "The environment does not support enumeration of states"
         )
 
+    @property
```
josephdviviano, line 278: thank you

src/gfn/gym/helpers/preprocessors.py
```diff
@@ -35,7 +35,6 @@ def __init__(
         self,
         height: int,
         ndim: int,
-        get_states_indices: Callable[[States], TT["batch_shape", "input_dim"]],
```
josephdviviano, line 38: This one i'm not sure of - why did you feel compelled to remove it?

src/gfn/utils/__init__.py
josephdviviano, line 1: I don't mind this change but I'm wondering what your motivation was here?

tutorials/examples/train_hypergrid.py
```diff
@@ -199,6 +199,9 @@ def main(args):  # noqa: C901
                 env, objects_type=objects_type, capacity=args.replay_buffer_size
             )
 
+    # Move the gflownet to the GPU.
+    gflownet = gflownet.to(device_str)
```
josephdviviano, line 203: Nice catch :)

tutorials/examples/train_hypergrid_simple.py
```diff
@@ -49,16 +47,13 @@
...
-
-for i in (pbar := tqdm(range(n_iterations))):
+for i in (pbar := tqdm(range(1000))):
     trajectories = sampler.sample_trajectories(
         env,
-        n_trajectories=batch_size,
+        n_trajectories=16,
         save_logprobs=False,
         save_estimator_outputs=True,
-        epsilon=exploration_rate,
```
josephdviviano, line 61: Why did you remove this (and `batch_size` and `n_iterations`)?

josephdviviano (changes requested):
Great stuff! A few style / other questions (nothing major but good to get on the same page before approval).

src/gfn/gym/helpers/preprocessors.py
hyeok9855: Since,
1. It was not used!
2. The parent class `Preprocessor` does not have it as a parameter

src/gfn/utils/__init__.py
hyeok9855: Previously, `NeuralNet` and `Tabular` were imported like this:

```python
from utils import NeuralNet, Tabular
```

However, I found this a bit odd since `utils` contains several submodules, each with its own specific functionality. So, I thought it made more sense to import them with the submodule specified, like this:

```python
from utils.modules import NeuralNet, Tabular
```

tutorials/examples/train_hypergrid_simple.py
hyeok9855: Sorry, I didn't mean to include this change initially—it was unintentionally committed with the whole file. However, after taking another look, I think there’s room for improvement, like using argparse instead of relying on global variables. I’ll push this update in the next commits!

josephdviviano: Totally agree with that!!

src/gfn/utils/__init__.py
josephdviviano: Yes I see, makes sense!

What I'm beginning to think is we will need a whole `modules` directory. Containing both the core GFN modules, plus these auxiliary modules, because I feel like we are going to get more and more of them over time - and the distinction between `modules.py` and `utils/modules.py` isn't obvious to new users :)

src/gfn/gym/helpers/preprocessors.py
josephdviviano: Oh no!  Great catch if so. Can you ensure this is truly not needed across the entire library? I don't mind if we keep this as part of the current PR.

Screenshot 2024-10-28 125646

hyeok9855: I've checked them thoroughly. The change has no problem, and the other parts seem to have no problem.

josephdviviano: Thanks!

josephdviviano (approved):
lgtm!! thanks

PR #204, opened 2024-10-22
Drop torchtyping

younik:
This is another proposal for dropping torchtyping, and using asserts instead (notice this is a stricter approach, as torch typing doesn't fail at runtime if shapes are not respected).

There are few shapes which are the ones typed with torchtyping, see comments.


src/gfn/containers/trajectories.py
```diff
@@ -90,17 +88,25 @@ def __init__(
...
+
+        if log_probs is not None:
+            assert log_probs.shape == (self.max_length, self.n_trajectories) and log_probs.dtype == torch.float
+        else:
+            log_probs = torch.full(size=(0, 0), fill_value=0, dtype=torch.float)
+        self.log_probs = log_probs        
+
         self.estimator_outputs = estimator_outputs
+        if self.estimator_outputs is not None:
+            # assert self.estimator_outputs.shape[:len(self.states.batch_shape)] == self.states.batch_shape TODO: check why fails
```
younik, line 108: This fails. In a test, `self.estimator_outputs.shape[:len(self.states.batch_shape)]` was `(8, 5)` while `self.states.batch_shape` was `(7, 5)`.

src/gfn/gym/helpers/box_utils.py
```diff
@@ -35,29 +34,28 @@ def __init__(
         self,
         delta: float,
         northeastern: bool,
-        centers: TT["n_states", 2],
-        mixture_logits: TT["n_states", "n_components"],
-        alpha: TT["n_states", "n_components"],
-        beta: TT["n_states", "n_components"],
+        centers: States,  # TODO: should probably be a tensor
```
younik, line 37: This was typed as Tensor, but it is a States. It makes more sense to be a Tensor, but for simplicity, I didn't change it in this PR.

src/gfn/gym/helpers/box_utils.py
```diff
@@ -550,6 +568,7 @@ def forward(
             desired_out[..., 1 + self._n_comp_max :]
         )
 
+        # assert desired_out.shape == batch_shape + (1 + 5 * self.n_components,)  # TODO: check why this fails
```
younik, line 571: This shape check fails. Not sure if the computation is correct?

src/gfn/samplers.py
```diff
@@ -84,10 +83,11 @@ def sample_actions(
             log_probs = None
 
         actions = env.actions_from_tensor(actions)
-
         if not save_estimator_outputs:
             estimator_output = None
 
+        assert log_probs is None or log_probs.shape == actions.batch_shape
+        # assert estimator_output is None or estimator_output.shape == actions.batch_shape  TODO: check expected shape
```
younik, line 90: This also fails. In a test, `estimator_output.shape` was (5, 3), while `actions.batch_shape` was (5, 2)

src/gfn/containers/trajectories.py
josephdviviano: That should not happen. We should investigate why the test fails. On my machine all tests pass... unless I misunderstand you?

src/gfn/gym/helpers/box_utils.py
josephdviviano: I would need to verify on my side, but all tests complete ok on my side. I find it as bit confusing, because we test the scripts as well.

Is it possible you introduced a bug?

src/gfn/samplers.py
josephdviviano: How are you producing these failures?

josephdviviano (changes requested):
Please see my notes from the closed PR.

src/gfn/samplers.py
younik: At the moment, I commented the failing asserts, so everything runs fine. However, if you uncomment the asserts and run tests they fail.

The intention for this PR was to leave the logic unchanged, so I should have not introduced a bug, but I might have overseen something

younik:
I added the shape in the documentation for each function that was using torchtyping @josephdviviano

src/gfn/samplers.py
josephdviviano: OK, thanks!

josephdviviano (approved):

josephdviviano (approved):
Really fantastic work Omar, thank you!

PR #207, opened 2024-10-28
Fix formatting

younik:
Changes after running `pre-commit run --all-files`, which fixes `black`, `flake8`, `autoflake` and `isort`.

(Another step towards https://github.com/GFNOrg/torchgfn/pull/193 to make easier focusing on pyright issues)


tutorials/examples/train_ising.py
```diff
@@ -81,10 +78,10 @@ def ising_n_to_ij(L, n):
     optimizer = torch.optim.Adam(gflownet.parameters(), lr=1e-3)
 
     # Learning
-    visited_terminating_states = env.States.from_batch_shape((0,))
+    env.States.from_batch_shape((0,))
```
josephdviviano, line 81: Does this do anything? It should be assigned.

Did you test all of the scripts (see `examples/test_scripts.py`)?

tutorials/examples/train_ising.py
```diff
@@ -26,9 +26,6 @@ def main(args):
 
     n_hidden = 2
     acc_fn = "relu"
-    lr = 0.001
-    lr_Z = 0.01
-    validation_samples = 1000
```
josephdviviano, line 31: Don't do this please. Variables should be named somehow to make it easier for the reader. I don't like magic numbers hanging around in scripts -- it requires more work to understand what they are for.

josephdviviano (changes requested):
Great - just one or two small questions :)

tutorials/examples/train_ising.py
younik: These variables are not used

Only `lr` could be inserted at line 73, but it has a different value, and I believe it is clearer as it is:
```python
optimizer = torch.optim.Adam(gflownet.parameters(), lr=1e-3)
```

tutorials/examples/train_ising.py
younik: `visited_terminating_states` is not used. Actually, the line itself is useless, so I should remove it instead(?)
Tests are working fine in both cases

younik: Actually this line was removed in https://github.com/GFNOrg/torchgfn/pull/203

tutorials/examples/train_ising.py
younik: Actually, https://github.com/GFNOrg/torchgfn/pull/203 inserted lr in that line
(what are the odds)

josephdviviano: OK - I must have lost track - thanks for following up on this.

tutorials/examples/train_ising.py
josephdviviano: OK makes sense to remove it then! Thanks

josephdviviano (approved):
lgtm!

PR #209, opened 2024-10-31
Add pre-commit workflow & fix remaining pyright issues

younik:
Aș per title, this PR adds the pre-commit workflow & fixes the remaining typing issues

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v4
         with:
-          python-version: '3.11'
+          python-version: '3.10'
       - run: pip install .[all]
       - run: pre-commit --version
       - run: pre-commit run --all-files
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,20 @@
+# https://pre-commit.com
+# This GitHub Action assumes that the repo contains a valid .pre-commit-config.yaml file.
+---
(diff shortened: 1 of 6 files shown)


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,20 @@
...
+  contents: read
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
```
josephdviviano, line 17: unlikely to be an issue but the codebase requires >=3.10 so maybe we should use that here.

src/gfn/containers/replay_buffer.py
```diff
@@ -180,6 +180,12 @@ def add(self, training_objects: Transitions | Trajectories | tuple[States]):
 
         # Our buffer is full and we will prioritize diverse, high reward additions.
         else:
+            if (
+                self.training_objects.log_rewards is None
+                or training_objects.log_rewards is None
+            ):
+                raise ValueError("log_rewards must be defined for prioritized replay.")
```
josephdviviano, line 187: good catch thank you

josephdviviano (changes requested):
LGTM! Just one comment re python versions. Thank you!

.github/workflows/pre-commit.yml
younik: Changed to use Python 3.10

josephdviviano (approved):
LGTM!

PR #212, opened 2024-11-08
Fix notebooks

josephdviviano:
Fixes the smiley and simple continuous notebook to properly use the current `torchgfn` API.


younik (commented):
LGTM! I have reproduced the output on my laptop, which is good.

I have a couple of questions. Unfortunately, I cannot comment on [diff](https://github.blog/changelog/2023-03-01-feature-preview-rich-jupyter-notebook-diffs/).

```diff
env = LineEnvironment(
-   mus=[-2, 4, 6, 10],
+   mus=[-3, 4, 6, 10],
    variances=[0.2, 0.4, 1, 0.2],
    n_sd=4.5,
    init_value=0
)
```
What is the rationale for this change (cell 16, continuous line tutorial)?


```diff
-def make_States_class(self) -> type[DiscreteStates]:
-  ...
+# def make_States_class(self) -> type[DiscreteStates]:
+#  ...
```
In the smiley tutorial, you commented out this function, as it is not necessary... can we just delete it?

josephdviviano:
Good catch. I'm not sure why, but leaving the value at -2 created a
plotting error. I couldn't figure out how to fix the plotting error and the
number itself is inconsequential.

I'll remove the commented out bits.

Joseph Viviano
viviano.ca

alip67:
I reproduced the output on colab, which is good. A few Suggestion: 
 1- Add !pip install trochgfn in the start of the Torchgfn section. 
 2- change the  `env.mask_invalid_actions` in the `FacesEnv` to the `self.mask_invalid_actions`
3- Add a colab link for all of the notebooks

josephdviviano:
Thanks good catch @alip67 RE: `env.` vs `self.`. OTOH, I don't really want to support colab, these notebooks are designed to be run on your local machine.

PR #215, opened 2024-11-13
Scalar estimators allow for the reduction over many output values (i.…

josephdviviano:
…e., the output of the nn.Module does not need to be a scalar, because the Estimator will apply a reduction to the final output if required).

I am not in love with the current code organization of `modules.py` -- there is some duplication, but I am thinking that a bigger refactoring effort might be en route and perhaps we should wait to optimize. Open to feedback on this!

### DIFF
--- a/src/gfn/modules.py
+++ b/src/gfn/modules.py
@@ -9,7 +9,6 @@ from gfn.preprocessors import IdentityPreprocessor, Preprocessor
 from gfn.states import DiscreteStates, States
 from gfn.utils.distributions import UnsqueezedCategorical
 
-
 REDUCTION_FXNS = {
     "mean": torch.mean,
     "sum": torch.sum,
--- a/src/gfn/modules.py
+++ b/src/gfn/modules.py
@@ -142,7 +142,7 @@ class GFNModule(ABC, nn.Module):
 
 
 class ScalarEstimator(GFNModule):
-    r"""Class for estimating scalars such as LogZ.
+    r"""Class for estimating scalars such as LogZ or state flow functions of DB/SubTB.
 
     Training a GFlowNet requires sometimes requires the estimation of precise scalar
     values, such as the partition function of flows on the DAG. This Estimator is
@@ -371,7 +371,7 @@ class ConditionalDiscretePolicyEstimator(DiscretePolicyEstimator):
 
 
 class ConditionalScalarEstimator(ConditionalDiscretePolicyEstimator):
-    r"""Class for conditionally estimating scalars such as LogZ.
+    r"""Class for conditionally estimating scalars (LogZ,  DB/SubTB state logF).
 
     Training a GFlowNet requires sometimes requires the estimation of precise scalar
     values, such as the partition function of flows on the DAG. In the case of a
--- a/src/gfn/modules.py
+++ b/src/gfn/modules.py
@@ -134,8 +134,70 @@ class GFNModule(ABC, nn.Module):
 
 
 class ScalarEstimator(GFNModule):
+    r"""Class for estimating scalars such as LogZ.
+
+    Training a GFlowNet requires sometimes requires the estimation of precise scalar
+    values, such as the partition function of flows on the DAG. This Estimator is
+    designed for those cases.
+
+    The function approximator used for `module` need not directly output a scalar. If
+    it does not, `reduction` will be used to aggregate the outputs of the module into
+    a single scalar.
+
+    Attributes:
+        preprocessor: Preprocessor object that transforms raw States objects to tensors
+            that can be used as input to the module. Optional, defaults to
+            `IdentityPreprocessor`.
+        module: The module to use. If the module is a Tabular module (from
+            `gfn.utils.modules`), then the environment preprocessor needs to be an
+            `EnumPreprocessor`.
+        preprocessor: Preprocessor from the environment.
+        _output_dim_is_checked: Flag for tracking whether the output dimenions of
+            the states (after being preprocessed and transformed by the modules) have
+            been verified.
+    """
+
+    def __init__(
+        self,
+        module: nn.Module,
+        preprocessor: Preprocessor | None = None,
+        is_backward: bool = False,
+        reduction: str = "mean",
+    ):
+        super().__init__(module, preprocessor, is_backward)
+        reduction_fxns = {
+            "mean": torch.mean,
+            "sum": torch.sum,
+            "prod": torch.prod,
+        }
+        assert reduction in reduction_fxns
+        self.reduction_fxn = reduction_fxns[reduction]
+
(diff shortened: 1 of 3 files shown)


src/gfn/modules.py
```diff
@@ -306,6 +394,13 @@ def __init__(
             preprocessor=preprocessor,
             is_backward=is_backward,
         )
+        reduction_fxns = {
+            "mean": torch.mean,
+            "sum": torch.sum,
+            "prod": torch.prod,
+        }
```
younik, line 401: here you can use the global constant, if you follow the previous comment

src/gfn/modules.py
```diff
@@ -134,9 +134,71 @@ def to_probability_distribution(
...
+        preprocessor: Preprocessor | None = None,
+        is_backward: bool = False,
+        reduction: str = "mean",
+    ):
+        super().__init__(module, preprocessor, is_backward)
+        reduction_fxns = {
+            "mean": torch.mean,
+            "sum": torch.sum,
+            "prod": torch.prod,
+        }
```
younik, line 172: this is constant, for convention it should go outside with upper case name

younik (approved):
Looks good! I added a comment for possible refactoring.

src/gfn/modules.py
```diff
@@ -134,9 +134,71 @@ def to_probability_distribution(
...
+        preprocessor: Preprocessor object that transforms raw States objects to tensors
+            that can be used as input to the module. Optional, defaults to
+            `IdentityPreprocessor`.
+        module: The module to use. If the module is a Tabular module (from
+            `gfn.utils.modules`), then the environment preprocessor needs to be an
+            `EnumPreprocessor`.
+        preprocessor: Preprocessor from the environment.
+        _output_dim_is_checked: Flag for tracking whether the output dimenions of
+            the states (after being preprocessed and transformed by the modules) have
+            been verified.
```
hyeok9855, line 157: This needs to be updated accordingly, e.g., add `is_backward` and `reduction` and remove `_output_dim_is_checked`.

src/gfn/modules.py
```diff
@@ -290,13 +352,39 @@ def forward(self, states: States, conditioning: torch.tensor) -> torch.Tensor:
...
+            `IdentityPreprocessor`.
+        module: The module to use. If the module is a Tabular module (from
+            `gfn.utils.modules`), then the environment preprocessor needs to be an
+            `EnumPreprocessor`.
+        preprocessor: Preprocessor from the environment.
+        reduction_fxn: the selected torch reduction operation.
+        _output_dim_is_checked: Flag for tracking whether the output dimensions of
+            the states (after being preprocessed and transformed by the modules) have
+            been verified.
+    """
```
hyeok9855, line 378: DITTO

src/gfn/modules.py
```diff
@@ -134,9 +134,71 @@ def to_probability_distribution(
 
 
 class ScalarEstimator(GFNModule):
+    r"""Class for estimating scalars such as LogZ.
```
hyeok9855, line 137: Note that `logZ` for unconditional TB is usually modeled with a single learnable parameter (`nn.Parameter`).
Should we consider modifying `ScalarEstimator` to support this kind of behavior?

src/gfn/samplers.py
```diff
@@ -35,7 +35,11 @@ def sample_actions(
         save_estimator_outputs: bool = False,
         save_logprobs: bool = True,
         **policy_kwargs: Any,
-    ) -> Tuple[Actions, torch.Tensor | None, torch.Tensor | None,]:
+    ) -> Tuple[
+        Actions,
+        torch.Tensor | None,
+        torch.Tensor | None,
+    ]:
```
hyeok9855, line 42: Removing the last `,` will make this one line.

src/gfn/modules.py
```diff
@@ -134,9 +134,71 @@ def to_probability_distribution(
...
+            "sum": torch.sum,
+            "prod": torch.prod,
+        }
+        assert reduction in reduction_fxns
+        self.reduction_fxn = reduction_fxns[reduction]
+
     def expected_output_dim(self) -> int:
         return 1
 
+    def forward(self, input: States | torch.Tensor) -> torch.Tensor:
```
hyeok9855, line 179: In which case is the input `torch.Tensor`?

hyeok9855 (commented):
Looks good! I left a few questions and comments below.

src/gfn/modules.py
josephdviviano: The GFNs themselves support this directly (you do not need to pass an estimator at all, instead you just pass a float for `Z`).

src/gfn/modules.py
josephdviviano: Yes I was looking at this and not entirely sure. It might be in the case of conditioning, where we currently don't have any sort of container, conditioning is done with a raw `Tensor`.

src/gfn/modules.py
josephdviviano: Thanks - I've done this

src/gfn/modules.py
hyeok9855: This comment is because of `such as LogZ` in the docstring!

src/gfn/modules.py
hyeok9855: Oh, it should be conditioning (e.g., conditional log Z(c)).
However, it might be a bit confusing whether to use `ConditionalScalarEstimator` or `ScalarEstimator` to model log Z(c).

josephdviviano: Well of note, ScalarEstimators are used for more than just logZ, but in this case, I see it like this:

- LogZ can be a single parameter.
- LogZ can be estimated using a neural network - in this case, the output of the network can actually be multiple items that are averaged over.
- LogZ can be conditionally estimated using a neural network  - in this case, the output of the network can actually be multiple items that are averaged over.

From an optimization POV, sometimes having logZ only be estimated by a single parameter can cause problems (i.e., the gradients push the number around a lot), so using a neural network helps.

I agree we could make it clearer though -- I am open to suggestions.

josephdviviano: `ConditionalScalarEstimator` is used to take in both the State and the Conditioning, i.e., it's a two-headed estimator. I think this is the normal conditioning case.

src/gfn/modules.py
josephdviviano: thanks :)

src/gfn/modules.py
josephdviviano: I'm not entirely sure what would be most clear here but I'm open to suggestions.

hyeok9855: Why not just state flow functions of DB/SubTB??

hyeok9855 (approved):
Approved!

Issue #205, opened 2024-10-22
Hypergrid.get_terminating_states_indices code


younik:
I believe this function is not correct; it should be without argument and use `self.sf` (?)

https://github.com/GFNOrg/torchgfn/blob/c62b5b03b7c2cb6c669f2a0623fbe6484995b943/src/gfn/gym/hypergrid.py#L149-L152

josephdviviano:
Yeah, you're right. It's also not used anywhere. I'll submit a patch to remove it.

josephdviviano:
Addressed here -- https://github.com/GFNOrg/torchgfn/pull/217

josephdviviano:
OK - actually, this is fine. It's supposed to return the indices of the final states as sampled. It's used in this file 

https://github.com/GFNOrg/torchgfn/blob/8259a2139ce9ccc22fb2832c45cb2ffb7846314c/src/gfn/utils/training.py#L20

to calculate the `l1_dist` with the true posterior. 

however I'm going to patch up one or two things to make this more obvious and make the tests more robust.

josephdviviano:
I've submitted a PR that hopefully addresses the confusion here.

josephdviviano:
There's no need for this function when it just calls `self.get_states_indices` -- we should simply directly call this.

josephdviviano:
Will follow up in new issue

PR #217, opened 2024-11-15
small fixes to seeding and hypergrid tests, docstring improvements to resolve confusion of `get_terminating_states_indices` purpose

josephdviviano:
see title

was inspired by this issue https://github.com/GFNOrg/torchgfn/issues/205

### DIFF
--- a/src/gfn/gym/hypergrid.py
+++ b/src/gfn/gym/hypergrid.py
@@ -178,7 +178,7 @@ class HyperGrid(DiscreteEnv):
 
         Returns the indices of the terminating states in the canonical ordering as a tensor of shape `batch_shape`.
         """
-        return self.get_states_indices(states)
+        return self.get_states_indices(self.all_states)
 
     @property
     def n_states(self) -> int:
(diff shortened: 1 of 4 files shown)


src/gfn/gym/hypergrid.py
```diff
@@ -86,6 +86,7 @@ def update_masks(self, states: type[DiscreteStates]) -> None:
         """Update the masks based on the current states."""
         # Not allowed to take any action beyond the environment height, but
         # allow early termination.
+        # TODO: do we need to handle the conditional case here?
```
hyeok9855, line 89: I don't think so. We may need another EnvBase that incorporates conditions.

hyeok9855 (approved):
LGTM!

younik (approved):

src/gfn/gym/hypergrid.py
josephdviviano: ConditionalEnvs are interesting. I should actually do this ASAP so we have a working conditional GFN example.

PR #224, opened 2024-12-11
Add warm-up functionality with tensor to trajectory helper functions

alexandrelarouche:
I'm hoping to get your insight on how to make this better. Some parts of the code are sketchy and have been highlighted with a WARNING tag in the comments.

### DIFF
--- a/src/gfn/utils/training.py
+++ b/src/gfn/utils/training.py
@@ -9,7 +9,7 @@ from gfn.env import DiscreteEnv, Env
 from gfn.gflownet import GFlowNet, TBGFlowNet
 from gfn.gflownet.base import PFBasedGFlowNet
 from gfn.samplers import Trajectories
-from gfn.states import States, stack_states
+from gfn.states import States
 
 
 def get_terminating_state_dist_pmf(env: Env, states: States) -> torch.Tensor:
@@ -91,6 +91,7 @@ def states_actions_tns_to_traj(
     states_tns: torch.Tensor,
     actions_tns: torch.Tensor,
     env: DiscreteEnv,
+    conditioning: torch.Tensor | None = None,
 ) -> Trajectories:
     """
     This utility function helps integrate external data (e.g. expert demonstrations)
@@ -101,6 +102,7 @@ def states_actions_tns_to_traj(
         states_tns: Tensor of shape [traj_len, *state_shape] containing states for a single trajectory
         actions_tns: Tensor of shape [traj_len] containing discrete action indices
         env: The discrete environment that defines the state/action spaces
+        conditioning: Tensor of shape [traj_len, *conditioning_shape] containing states for a single trajectory
 
     Returns:
         Trajectories: A Trajectories object containing the converted states and actions
@@ -130,7 +132,7 @@ def states_actions_tns_to_traj(
     # stack is a class method, so actions[0] is just to access a class instance and is not particularly relevant
     actions = actions[0].stack(actions)
     log_rewards = env.log_reward(states[-2])
-    states = stack_states(states)
+    states = states[0].stack_states(states)
     when_is_done = torch.tensor([len(states_tns) - 1])
 
     log_probs = None
@@ -139,6 +141,7 @@ def states_actions_tns_to_traj(
     trajectory = Trajectories(
         env,
         states,
+        conditioning,
         actions,
         log_rewards=log_rewards,
         when_is_done=when_is_done,
@@ -177,8 +180,10 @@ def warm_up(
         optimizer.zero_grad()
         if isinstance(gflownet, PFBasedGFlowNet):
             loss = gflownet.loss(
-                env, training_trajs, recalculate_all_logprobs=recalculate_all_logprobs
-            )
+                env,
+                training_trajs,
+                recalculate_all_logprobs=recalculate_all_logprobs,  # pyright: ignore
+            )  # pyright: ignore
         else:
             loss = gflownet.loss(env, training_trajs)
 
--- a/src/gfn/utils/training.py
+++ b/src/gfn/utils/training.py
@@ -1,15 +1,15 @@
 from collections import Counter
 from typing import Dict, Optional
 
+import torch
 from tqdm import trange
 
 from gfn.containers import ReplayBuffer
-from gfn.env import Env, DiscreteEnv
+from gfn.env import DiscreteEnv, Env
 from gfn.gflownet import GFlowNet, TBGFlowNet
 from gfn.gflownet.base import PFBasedGFlowNet
 from gfn.samplers import Trajectories
 from gfn.states import States, stack_states
-import torch
 
 
 def get_terminating_state_dist_pmf(env: Env, states: States) -> torch.Tensor:
--- a/src/gfn/utils/training.py
+++ b/src/gfn/utils/training.py
@@ -109,7 +109,7 @@ def states_actions_tns_to_traj(
         )
     if len(actions_tns.shape) != 1:
         raise ValueError(f"actions_tns must be 1D, got batch_shape {actions_tns.shape}")
-    if states_tns.shape[0] != actions_tns.shape[0]:
+    if states_tns.shape[0] != actions_tns.shape[0] + 1:
         raise ValueError(
             f"states and actions must have same trajectory length, got "
             f"states: {states_tns.shape[0]}, actions: {actions_tns.shape[0]}"
@@ -126,12 +126,9 @@ def states_actions_tns_to_traj(
     states = stack_states(states)
     when_is_done = torch.tensor([len(states_tns) - 1])
 
-    # WARNING: This is sketchy. Create dummy values to avoid indexing / batch shape errors.
-    # WARNING: Assumes gfn.loss() uses recalculate_all_logprobs=True (thus only PFBasedGFlowNet are supported right now)!!
-    # WARNING: To reviewers: Can we bypass needing to define this?
-    log_probs = torch.full(size=(len(actions), 1), fill_value=0, dtype=torch.float)
+    log_probs = None
+    estimator_outputs = None
 
-    estimator_outputs = torch.zeros((len(actions), 1, env.n_actions))
     trajectory = Trajectories(
         env,
         states,
--- a/src/gfn/utils/training.py
+++ b/src/gfn/utils/training.py
@@ -1,11 +1,14 @@
 from collections import Counter
 from typing import Dict, Optional
 
-import torch
+from tqdm import trange
 
-from gfn.env import Env
+from gfn.containers import ReplayBuffer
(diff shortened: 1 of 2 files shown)


saleml:
Thank you for the PR. Could you please elaborate a little bit more on it? What use-case are you targeting? Where do you use the new functions? Is there a way to test them and see their effects in the repo?

Thanks

saleml:
The states_actions_tns_to_traj function needs better input validation and documentation. Here's how I would modify it:

```
# Validate input shapes
    if states_tns.shape[1:] != env.state_shape:
        raise ValueError(
            f"states_tns state dimensions must match env.state_shape {env.state_shape}, "
            f"got shape {states_tns.shape[1:]}"
        )
    if len(actions_tns.shape) != 1:
        raise ValueError(f"actions_tns must be 1D, got batch_shape {actions_tns.shape}")
    if states_tns.shape[0] != actions_tns.shape[0]:
        raise ValueError(
            f"states and actions must have same trajectory length, got "
            f"states: {states_tns.shape[0]}, actions: {actions_tns.shape[0]}"
        )

    # ... rest of the code ...
    ```
    
    Possible docstrign to add:
    ```
        """Convert external state-action tensors into a Trajectories object.
    
    This utility function helps integrate external data (e.g. expert demonstrations) 
    into the GFlowNet framework by converting raw tensors into proper Trajectories objects.
    
    Args:
        states_tns: Tensor of shape [traj_len, *state_shape] containing states for a single trajectory
        actions_tns: Tensor of shape [traj_len] containing discrete action indices
        env: The discrete environment that defines the state/action spaces
        
    Returns:
        Trajectories: A Trajectories object containing the converted states and actions
        
    Raises:
        ValueError: If tensor shapes are invalid or inconsistent
    """
    ```
    
    For the warm_up, a docstring would be appreciated. I am not sure why `gfn.loss` admits an extra argument for TBloss. I will investigate it.

File: src/gfn/utils/training.py [hunk 1]
```diff
@@ -74,3 +77,69 @@ def validate(
...
+    env: Env,
+    n_steps: int,
+    batch_size: int,
+    recalculate_all_logprobs=True,
+):
+    t = trange(n_steps, desc="Bar desc", leave=True)
+    for epoch in t:
+        training_trajs = replay_buf.sample(batch_size)
+        optimizer.zero_grad()
+        if isinstance(gfn, TBGFlowNet):
```
saleml, line 136: with #231 , this could be changed to a cleaner test (if it's a PFBasedGFlowNet)

Quoding: Yes! Seeing your commit, I think this would be cleaner.

src/gfn/utils/training.py
```diff
@@ -74,3 +78,109 @@ def validate(
...
+    # stack is a class method, so actions[0] is just to access a class instance and is not particularly relevant
+    actions = actions[0].stack(actions)
+    log_rewards = env.log_reward(states[-2])
+    states = stack_states(states)
+    when_is_done = torch.tensor([len(states_tns) - 1])
+
+    # WARNING: This is sketchy. Create dummy values to avoid indexing / batch shape errors.
+    # WARNING: Assumes gfn.loss() uses recalculate_all_logprobs=True (thus only PFBasedGFlowNet are supported right now)!!
+    # WARNING: To reviewers: Can we bypass needing to define this?
+    log_probs = torch.full(size=(len(actions), 1), fill_value=0, dtype=torch.float)
```
josephdviviano, line 132: log_probs can be `None` - which will trigger a recalculate downstream (or, if it doesn't, we should fix that).

josephdviviano: In other words you can remove `129-132`.

src/gfn/utils/training.py
```diff
@@ -74,3 +78,109 @@ def validate(
...
+
+    estimator_outputs = torch.zeros((len(actions), 1, env.n_actions))
+    trajectory = Trajectories(
+        env,
+        states,
+        actions,
+        log_rewards=log_rewards,
+        when_is_done=when_is_done,
+        log_probs=log_probs,
```
josephdviviano, line 141: `log_probs=None`

src/gfn/utils/training.py
```diff
@@ -74,3 +78,109 @@ def validate(
...
+            f"states and actions must have same trajectory length, got "
+            f"states: {states_tns.shape[0]}, actions: {actions_tns.shape[0]}"
+        )
+
+    states = [env.states_from_tensor(s.unsqueeze(0)) for s in states_tns]
+    actions = [
+        env.actions_from_tensor(a.unsqueeze(0).unsqueeze(0)) for a in actions_tns
+    ]
+
+    # stack is a class method, so actions[0] is just to access a class instance and is not particularly relevant
```
josephdviviano, line 123: really appreciate this comment

josephdviviano (changes requested):
Hey, first I want to apologize for taking so long to review this. I hit a bit of a lull over Dec / early Jan and have been playing catchup.

This is a really nice PR, and a feature I'd be excited to use myself in some of the applications I've been looking at. My only request revolves around the use of the dummy `log_probs` - if our library is working properly, it should function as intended using `log_probs=None`, and if not, we should fix the downstream elements if they're misbehaving, because this is the intended use of the `Trajectories` container.

Awesome contribution, thank you very much!

src/gfn/utils/training.py
Quoding: Ok! I will do some testing this week to see if it behaves as I expect before removing.

alexandrelarouche:
@josephdviviano  After testing by sending `log_probs = None`, it seems like it is working. However, on any `Trajectories.extend()` call, I get the following error:
```
Traceback (most recent call last):
  File "/Users/quoding/Documents/PhD/gfn-explain/scripts/experiments/knapsack/basic_gfn.py", line 225, in main
    replay_buf.add(trajectories)
  File "/Users/quoding/.pyenv/versions/gfn_tf/lib/python3.10/site-packages/gfn/containers/replay_buffer.py", line 177, in add
    self._add_objs(training_objects)
  File "/Users/quoding/.pyenv/versions/gfn_tf/lib/python3.10/site-packages/gfn/containers/replay_buffer.py", line 149, in _add_objs
    self.training_objects.extend(training_objects)
  File "/Users/quoding/.pyenv/versions/gfn_tf/lib/python3.10/site-packages/gfn/containers/trajectories.py", line 260, in extend
    assert self.log_probs.shape == self.actions.batch_shape
AssertionError
```

Therefore, it seems that doing things this way is problematic with replay buffers, among other things.

I noticed that the assert statement has a TODO comment over it saying it could be remove? Was that one of the reasons why?

josephdviviano:
Ok great thanks for testing. We can patch up the downstream elements that
are complaining, this isn't expected behaviour.

Joseph (Mobile)

saleml:
Thank you @alexandrelarouche for the PR. I vote to merge it (once isort is fixed). Could you please run `pre-commit run --all` before committing? (you can skip the pytest checks locally, as they are long.

josephdviviano:
@alexandrelarouche can you post a minimal example I can use to replicate your error?

alexandrelarouche:
Yes. Here is an MRE which fails on the log_rewards shape (instead of the log_probs, previously). It seems like the same chunk of code is responsible.

MRE:
```python
from gfn.containers.replay_buffer import ReplayBuffer
from gfn.env import Env
from gfn.gym.hypergrid import HyperGrid
from gfn.states import States
from gfn.utils.training import states_actions_tns_to_traj, warm_up
import torch


if __name__ == "__main__":
    env = HyperGrid(2, 4)
    states = torch.tensor([[0, 0], [0, 1], [0, 2], [-1, -1]])
    actions = torch.tensor([1, 1, 2])
    replay_buffer = ReplayBuffer(env, "trajectories")
    trajs = states_actions_tns_to_traj(states, actions, env)

    replay_buffer.add(trajs)  # Errors happen here
```

Error:
```python
Traceback (most recent call last):
  File "/Users/quoding/Documents/Code/torchgfn/scripts.py", line 16, in <module>
    replay_buffer.add(trajs)  # Errors happen here
    ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/quoding/Documents/Code/torchgfn/src/gfn/containers/replay_buffer.py", line 74, in add
    self.training_objects.extend(training_objects)
  File "/Users/quoding/Documents/Code/torchgfn/src/gfn/containers/trajectories.py", line 286, in extend
    assert len(self.log_rewards) == self.actions.batch_shape[-1]
AssertionError
```

Responsible code:
```python
trajectories.py
L281
-------------------
        # Ensure log_probs/rewards are the correct dimensions. TODO: Remove?
        if self.log_probs.numel() > 0:
            assert self.log_probs.shape == self.actions.batch_shape

        if self.log_rewards is not None:
            assert len(self.log_rewards) == self.actions.batch_shape[-1]
```

alexandrelarouche (commented):
This seems good overall, but I am confused as to why we would want to obscure code further with class method `stack` for states while there is a function for this. The only reason I used the class method for actions was the absence of a `stack_actions` equivalent to `stack_states`.

src/gfn/utils/training.py
```diff
@@ -130,7 +132,7 @@ def states_actions_tns_to_traj(
     # stack is a class method, so actions[0] is just to access a class instance and is not particularly relevant
     actions = actions[0].stack(actions)
     log_rewards = env.log_reward(states[-2])
-    states = stack_states(states)
+    states = states[0].stack_states(states)
```
alexandrelarouche, line 135: The only reason I used the class method for actions was the absence of a `stack_actions` equivalent to `stack_states`. I think I would keep `stack_states` here, in other words.

josephdviviano: We moved the function `stack_states` into the class, which is why Salem made this change I think. The import no longer exists after updating (I believe he pulled from main).

src/gfn/utils/training.py
```diff
@@ -91,6 +91,7 @@ def states_actions_tns_to_traj(
     states_tns: torch.Tensor,
     actions_tns: torch.Tensor,
     env: DiscreteEnv,
+    conditioning: torch.Tensor | None = None,
```
josephdviviano, line 94: @saleml was the fix simply to account for conditioning? Otherwise I don't see what you did, other than changing the function call to a method call.

josephdviviano (approved):
Everything seems to be working on my side, though I don't truly understand what the fix was, in the interest of time I'll approve and we can move on.

saleml:
Hi @alexandrelarouche 

Apologies as I should have explained earlier.

Basically the codebase changed from the moment you started the PR and the moment we merged it, so I needed to merge master into your branch. The remaining problem was the need to check that `log_probs` is not only not `None`, but also not "empty".

PR #225, opened 2025-01-11
Fixpyproject

saleml:
Following poetry 2.0 release


younik (approved):
Tested locally, it works!

PR #230, opened 2025-01-21
Default behavior: recalculate logprobs

saleml:
Should solve #179 

The changes are explicit

PR #231, opened 2025-01-22
Standardize recalculate_all_logprobs parameter across PF-based GFlowNets

saleml:
This PR standardizes the handling of `recalculate_all_logprobs` parameter across all policy-based GFlowNet implementations (TB, DB, SubTB).

Changes:
- Added `recalculate_all_logprobs` parameter to `PFBasedGFlowNet.loss()` abstract method
- Added the parameter to all implementing classes (DB, TB, SubTB)
- Default behavior: use existing logprobs if available (when `recalculate_all_logprobs=False`)
- When `recalculate_all_logprobs=True`, force recalculation of logprobs even if they exist

This change:
- Makes behavior consistent across all PF-based GFlowNets
- Aligns with default sampling behavior where `save_logprobs=False` (introduced in #230)
- Removes special case handling in training loops where TBGFlowNet had different behavior than other implementations (identified in #224)
- Does not affect Flow Matching implementation which doesn't use logprobs

Context:
- #230 introduced `save_logprobs=False` as default during sampling to improve memory efficiency
- #224 highlighted inconsistency where only TBGFlowNet accepted `recalculate_all_logprobs` parameter
- This PR unifies the behavior: by default, use existing logprobs if they were saved, otherwise recalculate them

### DIFF
--- a/src/gfn/gflownet/base.py
+++ b/src/gfn/gflownet/base.py
@@ -71,7 +71,7 @@ class GFlowNet(ABC, nn.Module, Generic[TrainingSampleType]):
         """Converts trajectories to training samples. The type depends on the GFlowNet."""
 
     @abstractmethod
-    def loss(self, env: Env, training_objects: Any, **kwargs: Any):
+    def loss(self, env: Env, training_objects: Any):
         """Computes the loss given the training objects."""
 
 
@@ -113,11 +113,7 @@ class PFBasedGFlowNet(GFlowNet[TrainingSampleType], ABC):
 
     @abstractmethod
     def loss(
-        self,
-        env: Env,
-        training_objects: Any,
-        recalculate_all_logprobs: bool = False,
-        **kwargs: Any,
+        self, env: Env, training_objects: Any, recalculate_all_logprobs: bool = False
     ):
         """Computes the loss given the training objects.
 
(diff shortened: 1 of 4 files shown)


src/gfn/gflownet/base.py
```diff
@@ -71,17 +71,12 @@ def to_training_samples(self, trajectories: Trajectories) -> TrainingSampleType:
         """Converts trajectories to training samples. The type depends on the GFlowNet."""
 
     @abstractmethod
-    def loss(self, env: Env, training_objects: Any):
+    def loss(self, env: Env, training_objects: Any, **kwargs: Any):
```
younik, line 74: Why adding kwargs instead of `recalculate_all_logprobs`?

src/gfn/gflownet/base.py
```diff
@@ -116,6 +111,24 @@ def pf_pb_named_parameters(self):
     def pf_pb_parameters(self):
         return [v for k, v in self.named_parameters() if "pb" in k or "pf" in k]
 
+    @abstractmethod
+    def loss(
+        self,
+        env: Env,
+        training_objects: Any,
+        recalculate_all_logprobs: bool = False,
+        **kwargs: Any,
```
younik, line 120: Do we need this kwargs?

src/gfn/gflownet/base.py
saleml: not needed indeed

josephdviviano (approved):
lgtm - sorry I thought I already reviewed this one.

PR #233, opened 2025-01-24
Fix reverse_backward_trajectories and LocalSearchSampler for continuous case

hyeok9855:
Fix compatibility issue in `reverse_backward_trajectories` and `LocalSearchSampler` with continuous envs.

I tested Box env, and it works fine, although the performance of TB + LS is slightly worse than vanilla TB.

I've checked many times, and I believe the slightly inferior performance is not due to an implementation error but to its overexploitation behavior.


hyeok9855:
Pre-commit is working on (the test takes some time on my PC...). Will be updated soon!

josephdviviano (approved):
lgtm!

PR #234, opened 2025-01-24
Update pre-commit configuration: enforce pyright and optimize local testing

saleml:
Changes:

- Added pyright with proper project configuration
- Added type: ignore comments for current type issues (to be resolved later)
- Modified local pre-commit test scope to only run tests in testing/ folder
- Preserved full test suite requirement in GitHub CI

Note: While some type issues are temporarily ignored, they will need to be addressed in future PRs. The change maintains code quality while improving developer experience through faster local commits.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,38 +1,38 @@
 ---
-default_language_version:
-  node: 15.1.0
 repos:
-  - repo: https://github.com/python/black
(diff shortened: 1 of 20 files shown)


josephdviviano (approved):
lgtm -- thanks for your attention to detail here

younik (commented):
Good catch!

younik (approved):

Issue #235, opened 2025-01-24
Make pyright happy


saleml:
Following #234, where pyright is fixed, but `# pyright: ignore`s are introduced all over, we need to go through the many dozen ignored lines, and fix them.

This will probably take more than one PR, but in the long run, this will make the codebase much more robust, as it will avoid accumulating errors.

saleml:
Fixed in #245

PR #244, opened 2025-02-17
Add Append MDP environment (Set Addition) and Perfect Binary Tree environment (bijection traj -> term states)

alexandrelarouche:
Hi,

I have implemented two simple gym environments which I find useful when experimenting with GFNs.

First one is simply an append MDP where there is a set of items, and any item can be added to the set at any time. Exit is allowed at any time except at the initial state. An instance may look like the following MDP (without the terminal state to lighten the figure).

image

The second environment is a [Perfect Binary Tree](https://en.wikipedia.org/wiki/Binary_tree#Types_of_binary_trees) MDP, where there is a bijection between trajectories and terminating states. Action 0 goes left, action 1 goes right. One could easily extend this environment to non-binary trees. We can discuss if it is necessary or interesting for torchgfn to have a non-binary extension environments. An instance of Perfect Binary Tree may look like:
image

### DIFF
--- a/src/gfn/gym/set_addition.py
+++ b/src/gfn/gym/set_addition.py
@@ -1,0 +1,70 @@
+from typing import Callable
+from gfn.env import Actions, DiscreteEnv, DiscreteStates
+import torch
+
+
+class SetAddition(DiscreteEnv):
+    """Append only MDP, similarly to what is described in Remark 8 of Shen et al. 2023
+    [Towards Understanding and Improving GFlowNet Training](https://proceedings.mlr.press/v202/shen23a.html)
+    """
+
+    def __init__(self, n_items: int, max_items: int, reward_fn: Callable):
+        self.n_items = n_items
+        self.reward_fn = reward_fn
+        self.max_traj_len = max_items
+        n_actions = n_items + 1
+        s0 = torch.zeros(n_items)
+        state_shape = (n_items,)
+
+        super().__init__(n_actions, s0, state_shape)
+
+    def get_states_indices(self, states: DiscreteStates):
+        states_raw = states.tensor
+
+        canonical_base = 2 ** torch.arange(
+            self.n_items - 1, -1, -1, device=states_raw.device
+        )
+        indices = (canonical_base * states_raw).sum(-1).long()
+        return indices
+
+    def update_masks(self, states: DiscreteStates) -> None:
+        trajs_that_must_end = states.tensor.sum(dim=1) >= self.max_traj_len
+        trajs_that_may_continue = states.tensor.sum(dim=1) < self.max_traj_len
+
+        states.forward_masks[trajs_that_may_continue, : self.n_items] = (
+            states.tensor[trajs_that_may_continue] == 0
+        )
+
+        # Disallow everything for trajs that must end
+        states.forward_masks[trajs_that_must_end, : self.n_items] = 0
+        states.forward_masks[..., -1] = 1  # Allow exit action
+
+        states.backward_masks[..., : self.n_items] = states.tensor != 0
+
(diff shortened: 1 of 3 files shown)


saleml:
Thank you for the PR @alexandrelarouche 
Could you please `pre-commit run --all` before you commit, then re-push once it succeeds?
For pyright, if you can't solve the issues, you can ignore the lines with a specific comment (that we already used throughout the codevase)

alexandrelarouche:
Hey @saleml,

My bad, I did run the pre-commit yesterday, but apparently I was negligent in noticing the pyright checks failed. This does open up an avenue of discussion though - it seems strange to require declaring a `States` type hint in a `DiscreteEnv` instead of a `DiscreteStates`. Is that something we could aim to fix or an inherent limitation of pyright's understanding of inheritance?

Thanks for being so pro-active on the repo 🙂 .

saleml:
>  Is that something we could aim to fix or an inherent limitation of pyright's understanding of inheritance?
This is definitely something that *can* and *should* be fixed. 

The first fix should be, indeed, to change all mentions of `States` to `DiscreteStates` in the `DiscreteEnv` class.

There is an open issue (#235) for this type of changes, and we're hoping a pedantic power-user of the library would some day appear and solve it. Perhaps LLMs can do that soon. Otherwise, when/if time permits, I'll do it... 


Regarding the two environments you added, are there some published benchmarks, or tests for them? 
If you look at the `test_scripts.py` file, we give some examples of minimal performances to reach for the implemented environments. This should future-proof the environments. I strongly recommend a similar test for each of your environments (of the type: if you run TB with BS 128 using this type of preprocessor, you would get this performance).

Do you think this is something feasible with the current code?

alexandrelarouche:
> Regarding the two environments you added, are there some published benchmarks, or tests for them?
If you look at the test_scripts.py file, we give some examples of minimal performances to reach for the implemented environments. This should future-proof the environments. I strongly recommend a similar test for each of your environments (of the type: if you run TB with BS 128 using this type of preprocessor, you would get this performance).
Do you think this is something feasible with the current code?

As of now, I do not know of specific benchmarks for these specific environments. I use these environment mostly on paper to study behavior, as they can easily map to the applicative settings I am interested in. I am certain they map to some other applicative scenarios that were studied before, but they offer an abstraction layer above them. As such, I can hardly provide a satisfying test that would show my implementation is correct.

However, as I use them in my own research, I can provide the tests once the papers are out. In the meantime, I understand that this may not be satisfactory for a PR, and we can leave this open if you wish.

src/gfn/gym/set_addition.py
```diff
@@ -0,0 +1,72 @@
+from typing import Callable
+
+import torch
+
+from gfn.env import Actions, DiscreteEnv, DiscreteStates
+
+
+class SetAddition(DiscreteEnv):
+    """Append only MDP, similarly to what is described in Remark 8 of Shen et al. 2023
```
saleml, line 9: Could you please add a one or two line explanation of this environment, e.g. states, actions... ?

saleml (commented):
Thank you for the PR. As you might have observed, there have been major chances in the master/main branch of the repo since you started this PR. Besides the minor comment on the code, could you:
- Merge master into your branch
- Add a test for each of your environments, in `test_scripts.py`, to ensure that the current gflownet losses work on your environment (you can hardcode a specific value for the L1 distance desired for example). You can take inspiration from the other tests in that file.
?
I'll merge your branch into master right after that

josephdviviano:
@alexandrelarouche I'll take this over. Thanks so much for the work!

saleml (approved):

PR #245, opened 2025-02-21
Complete Type Safety: Eliminating All Pyright Errors

saleml:
# 🎯 Complete Type Safety: Eliminating All Pyright Errors

## 🌟 Major Achievement
- **ZERO pyright errors** across the entire codebase! 🎉
- Stricter type checking with `reportOptionalMemberAccess` and `reportArgumentType` set to "error"
- Critical step towards a production-ready, enterprise-grade codebase

### 1. New Type-Safe Containers 📦
- Introduced `StatePairs[DiscreteStates]` for robust state pair handling
- Generic `ReplayBuffer[ContainerType]` implementation
- Type-safe container operations with proper generics

### 2. Enhanced Type Safety in Core Components ⚡
- Removed ALL `pyright: ignore` comments by fixing underlying issues
- Proper type casting for `DiscreteStates` in training examples
- Improved null handling in log probabilities and state operations
- Better type hints in preprocessors and estimators

### 3. Configuration & Quality Assurance 🛠️
- Extended pyright coverage to include `tutorials/examples/` and `testing/`
- Upgraded pre-commit hooks for stricter type checking
- More comprehensive type validation across the project

### Why This Matters
1. **Code Reliability**: Eliminates an entire class of runtime errors
2. **Developer Experience**: Better IDE support and code navigation
3. **Maintainability**: Easier to refactor and extend code with confidence
4. **Documentation**: Types serve as living documentation

### Strategic Timing ⏰
- Critical to merge before the next release
- Should precede the graph PR to ensure type safety foundation
- Sets the standard for future contributions

## 🔄 Next Steps
While this PR achieves complete pyright compliance, future improvements could include:
- Further generic type constraints
- Additional custom type guards
- More specific type narrowing
- Enhanced type documentation

## 🎓 Technical Details
```diff
+ Added reportOptionalMemberAccess = "error"
+ Added reportArgumentType = "error"
- Removed all pyright: ignore comments
+ Extended type checking coverage
```

## 🚀 Call to Action
This PR represents a crucial milestone in code quality. Merging it now will:
1. Set a strong foundation for future development
2. Prevent type-related technical debt
3. Ensure the upcoming graph PR maintains type safety standards

**Ready for immediate review and high-priority merge** 🔥


saleml:
This is ready for review/discussion. 
Thanks

src/gfn/gflownet/base.py
```diff
@@ -74,6 +74,27 @@ def to_training_samples(self, trajectories: Trajectories) -> TrainingSampleType:
...
+            env: The environment to compute the loss for
+            trajectories: The trajectories to compute the loss from
+
+        Returns:
+            torch.Tensor: The computed loss
+        """
+        training_samples = self.to_training_samples(trajectories)
+        if isinstance(self, PFBasedGFlowNet):
+            # We know this is safe because PFBasedGFlowNet's loss accepts these arguments
+            return self.loss(env, training_samples, recalculate_all_logprobs=True)
```
hyeok9855, line 95: This can be inefficient when the logprobs of on-policy trajectories are already assigned. What about raise a warning in such case?

hyeok9855 (commented):
This is insane. You're the best!

I'm still reviewing; this is review for the src/gfn/container and src/gfn/gflownet parts.

---

In src/gfn/gflownet line 16-21, you may miss this thing:
```
TrainingSampleType = TypeVar(
    "TrainingSampleType", bound=Union[Container, tuple[States, ...]]
)
```
I guess we only need `Container` now, right?

src/gfn/containers/__init__.py
```diff
@@ -1,3 +1,12 @@
...
 from .trajectories import Trajectories
 from .transitions import Transitions
+
+__all__ = [
+    "PrioritizedReplayBuffer",
+    "ReplayBuffer",
+    "StatePairs",
+    "Trajectories",
+    "Transitions",
+]
```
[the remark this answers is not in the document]
saleml: yes, thanks

src/gfn/containers/replay_buffer.py
```diff
@@ -1,127 +1,105 @@
...
+                raise ValueError(f"Unsupported type: {type(training_objects)}")
+
+    def add(self, training_objects: ContainerType) -> None:
         """Adds a training object to the buffer."""
-        terminating_states = None
-        if isinstance(training_objects, tuple):
-            assert self.objects_type == "states" and self.terminating_states is not None
-            training_objects, terminating_states = training_objects  # pyright: ignore
+        self.initialize(training_objects)
+        assert self.training_objects is not None
```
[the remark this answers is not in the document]
saleml: ok

src/gfn/containers/replay_buffer.py
```diff
@@ -130,88 +108,76 @@ class PrioritizedReplayBuffer(ReplayBuffer):
...
-        terminating_states = None
-        if isinstance(training_objects, tuple):
-            assert self.objects_type == "states" and self.terminating_states is not None
-            training_objects, terminating_states = training_objects  # pyright: ignore
+        assert self.training_objects.log_rewards is not None
+        ix = torch.argsort(self.training_objects.log_rewards)
+        self.training_objects = cast(ContainerType, self.training_objects[ix])
+        self.training_objects = cast(
+            ContainerType, self.training_objects[-self.capacity :]
+        )
```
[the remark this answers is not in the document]
saleml: I don't decide. It's `black` that decides :) 
Basically in pyproject.toml, we have a black config. My IDE (VS Code) has the black extension. Whenever I save a file, it automatically applies black to it

src/gfn/containers/replay_buffer.py
```diff
@@ -130,88 +108,76 @@ class PrioritizedReplayBuffer(ReplayBuffer):
...
 
+    def add(
+        self,
+        training_objects: ContainerType,
+    ):
+        """Adds a training object to the buffer."""
         to_add = len(training_objects)
```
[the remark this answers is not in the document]
saleml: ok

src/gfn/containers/state_pairs.py
```diff
@@ -0,0 +1,131 @@
...
+        )
+
+        # We can construct a new StatePairs with the same StateType
+        return StatePairs[StateType](
+            env=self.env,
+            intermediary_states=intermediary_states,
+            terminating_states=terminating_states,
+            intermediary_conditioning=intermediary_conditioning,
+            terminating_conditioning=terminating_conditioning,
+        )
```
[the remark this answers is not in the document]
saleml: Good point. The only time this is used is for 2 dimensional StatePairs. The batch dimension should still be the same for both `States` objects. Should we ensure the indexing is done right?

src/gfn/containers/trajectories.py
```diff
@@ -369,11 +368,13 @@ def to_transitions(self) -> Transitions:
...
-        log_probs = (
-            self.log_probs[~self.actions.is_dummy] if has_log_probs(self) else None
-        )
+        # Initialize log_probs as zeros if not available
+        if has_log_probs(self):
+            log_probs = self.log_probs[~self.actions.is_dummy]
+        else:
+            log_probs = torch.zeros(
+                actions.batch_shape[0], dtype=torch.float, device=actions.device
+            )
```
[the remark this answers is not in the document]
saleml: that's a good point, perhaps we should keep it None

src/gfn/containers/trajectories.py
```diff
@@ -392,21 +393,17 @@ def to_states(self) -> States:
...
+    def to_state_pairs(
         self,
-    ) -> Union[
-        Tuple[States, States, torch.Tensor, torch.Tensor],
-        Tuple[States, States, None, None],
-    ]:
-        """Returns all intermediate and terminating `States` from the trajectories.
-
-        This is useful for the flow matching loss, that requires its inputs to be distinguished.
+    ) -> StatePairs[DiscreteStates]:
```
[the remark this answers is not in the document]
saleml: true !

src/gfn/actions.py
```diff
@@ -2,7 +2,7 @@
 
 from abc import ABC
 from math import prod
-from typing import ClassVar, Sequence
+from typing import ClassVar, List, Sequence
```
hyeok9855, line 6: What about using only the python3.9 style type hinting (e.g., List -> list, Optional -> | None, etc.)

src/gfn/gflownet/base.py
saleml: yes, good point. thanks

src/gfn/gflownet/flow_matching.py
```diff
@@ -178,36 +179,25 @@ def reward_matching_loss(
     def loss(
         self,
         env: DiscreteEnv,
-        states_tuple: Union[
-            Tuple[DiscreteStates, DiscreteStates, torch.Tensor, torch.Tensor],
-            Tuple[DiscreteStates, DiscreteStates, None, None],
-        ],
+        states: StatePairs[DiscreteStates],
```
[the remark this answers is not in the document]
saleml: good point

src/gfn/actions.py
saleml: I don't have a strong opinion about this? For consistency with other types (that start with a capital letter), I prefer List over list. Is there a specific reason you prefer python3.9 style type hints ?

src/gfn/containers/replay_buffer.py
```diff
@@ -1,127 +1,105 @@
...
 
 import os
-from typing import TYPE_CHECKING, Literal
+from typing import Generic, TypeVar, cast
 
 import torch
 
+from gfn.containers.state_pairs import StatePairs
 from gfn.containers.trajectories import Trajectories
 from gfn.containers.transitions import Transitions
```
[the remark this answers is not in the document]
saleml: well actually no. Because it's in the same folder. So it's a circular import. __init__ of this folder requires ReplayBuffer

saleml:
> In src/gfn/gflownet line 16-21, you may miss this thing:
> 
> ```
> TrainingSampleType = TypeVar(
>     "TrainingSampleType", bound=Union[Container, tuple[States, ...]]
> )
> ```
> 
> I guess we only need `Container` now, right?

You are right! I have just changed it

src/gfn/containers/replay_buffer.py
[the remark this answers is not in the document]
hyeok9855: I actually meant that the two indexing operations can be done in a single line, but this is too minor; you can skip it.

src/gfn/containers/replay_buffer.py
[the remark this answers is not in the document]
hyeok9855: Oh, I see. My bad.

src/gfn/actions.py
hyeok9855: Just because it looks better, though I don’t feel too strongly about it.

src/gfn/containers/replay_buffer.py
```diff
@@ -1,127 +1,105 @@
...
-    def add(self, training_objects: Transitions | Trajectories | tuple[States]):
-        """Adds a training object to the buffer."""
-        terminating_states = None
-        if isinstance(training_objects, tuple):
-            assert self.objects_type == "states" and self.terminating_states is not None
-            training_objects, terminating_states = training_objects  # pyright: ignore
+        return 0 if self.training_objects is None else len(self.training_objects)
+
+    def initialize(self, training_objects: ContainerType) -> None:
+        """Initializes the buffer with a training object."""
```
hyeok9855, line 56: It's minor; What about "maybe_initialize" ?

testing/test_environments.py
```diff
@@ -91,35 +87,28 @@ def test_HyperGrid_fwd_step_with_preprocessors(
...
     for actions_list in passing_actions_lists:
         actions = env.actions_from_tensor(format_tensor(actions_list))
-        states = env._step(states, actions)  # pyright: ignore
+        states = env._step(states, actions)
 
     # Step 4 fails due an invalid input action.
     actions = env.actions_from_tensor(format_tensor(failing_actions_list))
     with pytest.raises(NonValidActionsError):
-        states = env._step(states, actions)  # pyright: ignore
+        states = env._step(states, actions)
```
hyeok9855, line 95: What about exchanging the name of `env.step` and `env._step`? Currently, `env._step` is the one that needs to be called externally (as seen in samplers.py), which seems unusual for something with a private-style naming convention.

The same goes for `env._backward_step`.

tutorials/examples/train_conditional.py
```diff
@@ -206,6 +202,7 @@ def train(env, gflownet, seed):
 
     print("+ Training Conditional {}!".format(type(gflownet)))
     for i in (pbar := tqdm(range(n_iterations))):
+        _ = i  # Used in progress bar
         conditioning = torch.rand((batch_size, 1))
```
hyeok9855, line 206: May I ask why this is needed?

testing/test_environments.py
saleml: Last year, we used to have `maskless_step` and `step`. 
We then changed them to `step` and `_step` respectively.

A user that defines their environment needs to define `step` only, and `_step` handles the masking for them.

When I don't work on the codebase for 1-2 months and go back to it, I agree that `_step` is confusing. What do you think of we change it to `safe_step`? (obviously, in a new enviornment, the user would still need to write `step` only).

@josephdviviano , your opinion here would be appreciated too. Thanks

src/gfn/samplers.py
```diff
@@ -706,35 +710,46 @@ def _combine_prev_and_recon_trajectories(  # noqa: C901
...
+                    use_metropolis_hastings
+                    and prev_trajectories_log_pb is not None
+                    and recon_trajectories_log_pb is not None
+                ):
+                    _new_trajectories_log_pb[:_n_prev, i] = prev_trajectories_log_pb[
+                        :_n_prev, i
+                    ]
+                    _new_trajectories_log_pb[_n_prev : _n_prev + _len_recon, i] = (
+                        recon_trajectories_log_pb[:_len_recon, i]
+                    )
```
hyeok9855, line 735: The repeated `is not None` checks could be refactored. I’ll take care of this and push a commit soon.

src/gfn/containers/state_pairs.py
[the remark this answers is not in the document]
hyeok9855: Oh, I see. But I think we still need right indexing for the replay buffer to work properly.

testing/test_environments.py
hyeok9855: I don’t really have a strong preference on this —`safe_step` seems to be fine also.

src/gfn/samplers.py
hyeok9855: Well, I found it is not trivial to avoid the repeated check...

src/gfn/containers/state_pairs.py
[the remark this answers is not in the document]
saleml: I'll make the docstring more explicit

src/gfn/containers/replay_buffer.py
saleml: Not sure I understand why the maybe
This surely initialized the replay buffer. Initially it's empty, and when `add` is called for the first time, this function is called, and it's not empty anymore. 
What did you have in mind?

tutorials/examples/train_conditional.py
saleml: Indeed. Not needed :) thanks

saleml:
Thank you for your commit @hyeok9855. I have addressed all your points, and left a question.

src/gfn/containers/replay_buffer.py
hyeok9855: I'm asking this because we call the `self.initialize` every time we call `self.add`. So, the `self.initialize` just returns `None` once the `self.training_objects` is initialized, right?
What about (in `self.add`) calling this method only when `self.training_objects` is `None`?

src/gfn/containers/state_pairs.py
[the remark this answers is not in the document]
hyeok9855: No, I mean this still raises the IndexError.
You can simply check by:
```bash
python tutorials/examples/train_hypergrid.py --loss FM --replay_buffer_size 100
```

src/gfn/containers/trajectories.py
```diff
@@ -369,11 +368,11 @@ def to_transitions(self) -> Transitions:
...
-        # FIXME: Transitions requires log_probs for initialization (see line 107 in transitions.py).
-        # Shouldn't we make sure that log_probs are always available?
-        log_probs = (
-            self.log_probs[~self.actions.is_dummy] if has_log_probs(self) else None
-        )
+        # Initialize log_probs None if not available
+        if has_log_probs(self):
+            log_probs = self.log_probs[~self.actions.is_dummy]
+        else:
+            log_probs = None
```
hyeok9855, line 375: Hey Salem, I think this is problematic. You can simply check the error by
```bash
python tutorials/examples/train_hypergrid.py --loss DB --replay_buffer_size 100
```

The error comes from the assertion [here](https://github.com/GFNOrg/torchgfn/blob/b6fb61316200e2a2ace164280486678c55d36ce2/src/gfn/containers/transitions.py#L107).

saleml: good catch. I have locally added some tests to catch such errors, and there were more. Fixing them now. Pushing soon

src/gfn/containers/transitions.py
```diff
@@ -104,10 +104,6 @@ def __init__(
             and self._log_rewards.dtype == torch.float
         )
         self.log_probs = log_probs if log_probs is not None else torch.zeros(0)
-        assert (
-            self.log_probs.shape == (self.n_transitions,)
-            and self.log_probs.dtype == torch.float
-        )
```
hyeok9855, line 110: Instead of removing this, what about check two the cases like this:
```
        assert (
            (self.log_probs.shape == (self.n_transitions,) or len(self.log_probs) == 0)
            and self.log_probs.dtype == torch.float
        )
```

src/gfn/containers/state_pairs.py
```diff
@@ -60,7 +60,7 @@ def __init__(
         self.terminating_conditioning = terminating_conditioning
 
     def __len__(self) -> int:
-        return len(self.intermediary_states) + len(self.terminating_states)
+        return min(len(self.intermediary_states), len(self.terminating_states))
```
hyeok9855, line 77: This approach feels a bit like a workaround. Using the sum of the two lengths seems more reasonable to me. 

It looks like we might need to refactor this a bit more. For example, instead of storing `intermediary_states` and `terminating_states` in separate variables, we could combine them into one and use two sets of indices to track whether a state is intermediate or terminating. 

If you agree, I'll go ahead and create an issue for this. Let me know if you have any better suggestions!"

hyeok9855 (approved):
I approved with one concern; see below.

I really appreciate this PR, @saleml !!

src/gfn/containers/transitions.py
saleml: That's a good point. I have added it.

src/gfn/containers/state_pairs.py
saleml: I agree it's a bit hacky. This is only because we *have* to have a `len` function in
```
def sample(self, n_samples: int) -> Container:
        """Samples a subset of the container."""
        return self[torch.randperm(len(self))[:n_samples]]
```
in `containers/base.py`.
Please go ahead and raise the issue. Thanks for your review.

hyeok9855: #247

testing/test_environments.py
josephdviviano: One option might be to call this method `env._base_step` -- but I think we should keep the `_` which denotes to the user of the library "you shouldn't call this method unless you really know what you're doing".

josephdviviano (commented):
Partial review of the core features. I didn't check everything. Thanks very much to @hyeok9855 for his thorough review. I do however have some important comments to address.

.github/workflows/pre-commit.yml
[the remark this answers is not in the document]
saleml: this was a duplicate of `ci.py`. Everything is still tested for on github (e.g., this PR)

.github/workflows/python-package-conda.yml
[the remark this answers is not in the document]
saleml: ditto

pyproject.toml
```diff
@@ -25,11 +25,12 @@ classifiers = [
 einops = ">=0.6.1"
 numpy = ">=1.21.2"
 python = "^3.10"
-torch = ">=1.9.0"
+torch = "==2.6.0"
```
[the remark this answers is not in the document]
saleml: ok

pyproject.toml
```diff
@@ -25,11 +25,12 @@ classifiers = [
...
 
 # dev dependencies.
 black = { version = "24.3", optional = true }
 flake8 = { version = "*", optional = true }
+pyright = {version = "*", optional = true}
```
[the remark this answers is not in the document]
saleml: ok

src/gfn/gflownet/base.py
saleml: we don't have that flag anymore. I improved the warning.

src/gfn/utils/modules.py
```diff
@@ -34,6 +34,7 @@ def __init__(
         self._output_dim = output_dim
 
         if trunk is None:
+            hidden_dim = hidden_dim or 256
```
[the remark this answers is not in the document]
saleml: yes true

saleml:
> One option might be to call this method env._base_step -- but I think we should keep the _ which denotes to the user of the library "you shouldn't call this method unless you really know what you're doing"

No strong opinion here. We can have this discussion outside this PR.

Thanks for your review. I have addressed your comments.

josephdviviano (commented):
Additional comment

PR #246, opened 2025-02-25
bitsequence environment

Idriss-Malek:


src/gfn/gym/bitSequence.py
```diff
@@ -0,0 +1,755 @@
...
+
+        Returns:
+            torch.Tensor: A tensor containing the unique mode sequences.
+
+        Raises:
+            ValueError: If the number of requested modes exceeds the number of possible unique sequences.
+
+        Notes:
+            - The method sets the random seed for both CPU and GPU (if available) to ensure reproducibility.
+            - The tensor H is initialized with predefined sequences if it is not already set. This default value is the one chosen in the TB objectie paper.
```
saleml, line 408: typo

src/gfn/gym/bitSequence.py
```diff
@@ -0,0 +1,755 @@
...
+        Returns:
+            BitSequenceStates: The new states after performing the backward step.
+        """
+        new_states = super(DiscreteEnv, self)._backward_step(states, actions)
+        assert isinstance(new_states, BitSequenceStates)
+        new_states.length = states.length - 1
+        self.update_masks(new_states)
+        return new_states
+
+    def make_modes_set(self, seed: int = 42) -> torch.Tensor:
```
saleml, line 393: Could you explain in one or two sentences in the docstring how the modes are defined from H?

src/gfn/gym/bitSequence.py
```diff
@@ -0,0 +1,755 @@
...
+        exit_action (torch.Tensor): A tensor representing the exit action.
+    """
+
+    def __init__(
+        self,
+        word_size: int = 4,
+        seq_size: int = 120,
+        n_modes: int = 60,
+        temperature: float = 1.0,
+        H: Optional[torch.Tensor] = None,
```
saleml, line 197: I believe the seed of make_modes_set should be part of the init function of the environment. As of now, the user doesn't have any way to set the seed, and it will always be 42. BTW can you use 0 for default seed, as in the rest of the database? Interestingly, whenever an LLM is tasked of generating random code, it uses seed 42

saleml (commented):
Thanks for the PR. Minor comments above!
Could you add a test to `test_scripts.py` that involve this new environment? It can be a few steps of SGD with TB loss or a loss of your choice, to ensure that it goes down, and matches some experiments you might have done with this environment. You could take inspiration from the other tests in the same file.

saleml (approved):
Thanks for the changes, and the great PR. This is a useful addition to the library.

Issue #255, opened 2025-03-06
To support `GraphStates`, maybe utilize `States.__repr__`?


saleml:
TODO: To support `GraphStates`, maybe utilize `States.__repr__`?

_Originally posted by @hyeok9855 in https://github.com/GFNOrg/torchgfn/pull/252#discussion_r1983724900_

PR #261, opened 2025-03-12
added graph representation, modified all state representations, modified trajectory representations

josephdviviano:
addresses https://github.com/GFNOrg/torchgfn/issues/255

note that the graph representations are not legible. This could be addressed easily in a follow up PR by modifying the graphstate `__repr__` - but it isn't clear to me how to make this easy to read.

### DIFF
--- a/src/gfn/containers/trajectories.py
+++ b/src/gfn/containers/trajectories.py
@@ -120,7 +120,7 @@ class Trajectories(Container):
 
     def __repr__(self) -> str:
         trajectories_representation = ""
-        n_traj_to_print = 10
+        n_traj_to_print = min(10, self.n_trajectories)
 
         if isinstance(self.states, GraphStates):
             for i in range(n_traj_to_print):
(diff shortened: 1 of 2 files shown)


src/gfn/containers/trajectories.py
```diff
@@ -119,25 +119,26 @@ def __init__(
...
-                if self.is_backward and step.equal(self.env.s0):
-                    break
-                elif not self.is_backward and step.equal(self.env.sf):
-                    break
-            trajectories_representation += "-> ".join(one_traj_repr) + "\n"
+        n_traj_to_print = 10
+
+        if isinstance(self.states, GraphStates):
+            for i in range(n_traj_to_print):
+                trajectories_representation += str(self.states[..., i]) + "\n"
```
younik, line 127: This may fail if self.states has less than 10 trajectory

younik (approved):
Looks good, thanks!

src/gfn/containers/trajectories.py
hyeok9855: I fixed it

hyeok9855:
LGTM

Issue #248, opened 2025-02-28
`gflownet.base.loss_from_trajectories` always recalculates logprobs


josephdviviano:
Right now `gflownet.base.loss_from_trajectories` always recalculates logprobs, even when the training is on policy. This is wasteful. We currently throw a warning but this does not help the user make the code more efficient. We should be able to pass a  flag of some kind to resolve this in the case that the user receives the warning.

saleml:
Thank you for this issue. Please note that this is only relevant after #245 is merged, as `gflownet.base.loss_from_trajectories` does not exist in the current version.

josephdviviano:
Addressed in https://github.com/GFNOrg/torchgfn/pull/262

PR #262, opened 2025-03-13
Handle recalculate_all_logprobs consistently across the library, and throw warnings where appropriate

josephdviviano:
Addresses https://github.com/GFNOrg/torchgfn/issues/248 (with some other modifications):

- `recalculate_all_logprobs` is default `True` everywhere. This is wasteful when training on-policy, but is the safest default behaviour. When the user recalcs logprobs and the container has stored log_probs, a warning is thrown.
- `loss_from_trajectories` now accepts `recalulate_all_logprobs`.
- `FlowMatchingGFlowNet.loss()` now accepts `recalculate_all_logprobs`. This is for API consistency, because as it stands, the container for this loss does not store logprobs. If the user tries to set this, a warning is thrown.

### DIFF
--- a/src/gfn/containers/base.py
+++ b/src/gfn/containers/base.py
@@ -55,10 +55,10 @@ class Container(ABC):
             else:
                 raise ValueError(f"Unexpected {key} of type {type(val)}")
 
+    @property
+    def has_log_probs(self) -> bool:
+        """Returns True if the container has the log_probs attribute populated."""
+        if not hasattr(self, "log_probs"):
+            return False
 
-def has_log_probs(obj: Container):
-    """Returns True if the submitted container has the log_probs attribute populated."""
-    if not hasattr(obj, "log_probs"):
-        return False
-
-    return obj.log_probs is not None and obj.log_probs.nelement() > 0
+        return self.log_probs is not None and self.log_probs.nelement() > 0
--- a/src/gfn/containers/base.py
+++ b/src/gfn/containers/base.py
@@ -54,3 +54,11 @@ class Container(ABC):
                 self.__dict__[key] = torch.load(os.path.join(path, key + ".pt"))
             else:
                 raise ValueError(f"Unexpected {key} of type {type(val)}")
+
+
+def has_log_probs(obj: Container):
+    """Returns True if the submitted container has the log_probs attribute populated."""
+    if not hasattr(obj, "log_probs"):
+        return False
+
+    return obj.log_probs is not None and obj.log_probs.nelement() > 0
--- a/src/gfn/gflownet/base.py
+++ b/src/gfn/gflownet/base.py
@@ -10,7 +10,6 @@ from gfn.env import Env
 from gfn.modules import GFNModule
 from gfn.samplers import Sampler
 from gfn.states import States
-from gfn.utils.handlers import warn_about_recalculating_logprobs
 from gfn.utils.prob_calculations import get_trajectory_pfs_and_pbs
 
 TrainingSampleType = TypeVar("TrainingSampleType", bound=Container)
@@ -103,7 +102,6 @@ class GFlowNet(ABC, nn.Module, Generic[TrainingSampleType]):
             torch.Tensor: The computed loss
         """
         training_samples = self.to_training_samples(trajectories)
-        warn_about_recalculating_logprobs(trajectories, recalculate_all_logprobs)
 
         return self.loss(
             env,
--- a/src/gfn/gflownet/base.py
+++ b/src/gfn/gflownet/base.py
@@ -97,6 +97,8 @@ class GFlowNet(ABC, nn.Module, Generic[TrainingSampleType]):
         Args:
             env: The environment to compute the loss for
             trajectories: The trajectories to compute the loss from
+            recalculate_all_logprobs: If True, always recalculate logprobs even
+                if they exist. If False, use existing logprobs when available.
 
         Returns:
             torch.Tensor: The computed loss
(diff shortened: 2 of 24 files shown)


tutorials/examples/test_scripts.py
```diff
@@ -8,9 +8,14 @@
...
-from .train_discreteebm import main as train_discreteebm_main
-from .train_hypergrid import main as train_hypergrid_main
+try:
+    from train_box import main as train_box_main
+    from train_discreteebm import main as train_discreteebm_main
+    from train_hypergrid import main as train_hypergrid_main
+except ImportError:
+    from .train_box import main as train_box_main
+    from .train_discreteebm import main as train_discreteebm_main
+    from .train_hypergrid import main as train_hypergrid_main
```
hyeok9855, line 18: Do we need this?

src/gfn/gflownet/base.py
```diff
@@ -89,18 +103,13 @@ def loss_from_trajectories(
...
         training_samples = self.to_training_samples(trajectories)
-        if isinstance(self, PFBasedGFlowNet):
-            # Check if trajectories already have log_probs
-            if has_log_probs(trajectories):
-                warnings.warn(
-                    "Recalculating logprobs for trajectories that already have them. "
-                    "This may be inefficient for on-policy trajectories. "
-                    "If the training is done on-policy, you should call loss() directly with recalculate_all_logprobs=False instead of loss_from_trajectories()."
-                )
+        warn_about_recalculating_logprobs(trajectories, recalculate_all_logprobs)
```
hyeok9855, line 106: Maybe we don't need this line because we have it in each `self.loss`

hyeok9855 (approved):
LGTM. Thanks!

tutorials/examples/test_scripts.py
josephdviviano: it seems to be required for running pytest from the commandline. i agree it's not elegant

younik: The relative import works for me by running `pytest tutorials/examples/test_scripts.py `

src/gfn/utils/handlers.py
```diff
@@ -40,3 +43,16 @@ def is_callable_exception_handler(
             f"conditioning was passed but {target_name} is not callable: {type(target)}"
         )
         raise
+
+
+def warn_about_recalculating_logprobs(
```
saleml, line 48: This is a very useful warning. I'll be a bit pedantic here:
Sometimes, log_probs are stored, but they are not the log_probs that should be used for the loss (e.g. the training objects come from another gflownet). Can you make if more explicit, by saying. "If this is the case, you should instead ..."

src/gfn/containers/base.py
```diff
@@ -54,3 +54,11 @@ def load(self, path: str) -> None:
...
             else:
                 raise ValueError(f"Unexpected {key} of type {type(val)}")
+
+
+def has_log_probs(obj: Container):
+    """Returns True if the submitted container has the log_probs attribute populated."""
+    if not hasattr(obj, "log_probs"):
+        return False
+
+    return obj.log_probs is not None and obj.log_probs.nelement() > 0
```
saleml, line 64: I'm curious to understand why chose it to be outside the class. Can't it be an instance method?

src/gfn/gflownet/base.py
saleml: It seems that the user would be warned twice here. Right?

src/gfn/gflownet/base.py
```diff
@@ -70,11 +69,26 @@ def to_training_samples(self, trajectories: Trajectories) -> TrainingSampleType:
...
+            recalculate_all_logprobs: If True, always recalculate logprobs even
+                if they exist. If False, use existing logprobs when available.
+        """
 
     def loss_from_trajectories(
-        self, env: Env, trajectories: Trajectories
+        self,
+        env: Env,
+        trajectories: Trajectories,
+        recalculate_all_logprobs: bool = True,
```
saleml, line 91: Given that this is the default behavior now, should we change anything in the scripts?

tutorials/examples/test_scripts.py
saleml: Same as @younik. I also tried `pytest` from the directory, and it works fine. @josephdviviano, what command did not work for you ?

saleml (changes requested):
Thank you for the great PR. please see me comments.

src/gfn/gflownet/base.py
hyeok9855: Yes it seems so

tutorials/examples/test_scripts.py
josephdviviano: I must have done something wrong. Reverting this change.

src/gfn/gflownet/base.py
josephdviviano: Good catch, removed.

src/gfn/utils/handlers.py
josephdviviano: Pedantic is good! thanks!

src/gfn/containers/base.py
josephdviviano: I actually just moved this from another location in the code to avoid circular imports. You're right, it's even cleaner as a instance method of the base class.

src/gfn/gflownet/base.py
josephdviviano: Perhaps yes - I haven't looked at it yet.

saleml: Actually yes, it's important. Both in scripts and tests.
I'm on it...

saleml:
@josephdviviano , please look at my 2 commits.
If you agree with them, feel free to merge.

Issue #228, opened 2025-01-13
'States' object has no attribute 'shape'


philmar1:
Hi! 

I'm trying to build a simple autoregressive model to generate sequences of str using torchgfn 

I get the following error: 'States' object has no attribute 'shape' (line 390 from env.py in the __init__() of class DiscreteEnv. : assert s0.shape == state_shape)

I'm using torchgfn 1.1.1

saleml:
Hi @philmar1, looking at the main branch, it looks like `s0` is a `Tensor` object, and not a `States` object at initialization of `DiscreteEnv`. In your code, are you initializing with a tensor or with a `State` ?

josephdviviano:
@philmar1 sorry for the lag in my reply to this.

I just reproduced your bug. The fix is simple -- I'll put the appropriate handling into the library to avoid this bug. I'm going to submit a fix now (it will be available on `master` immediately.

```
dummy_action.shape
torch.Size([10])

action_shape
10

dummy_action.shape == (action_shape)
False

dummy_action.shape == (action_shape,)
True
```

josephdviviano:
To clarify, the easier way to do this would be (following `gym/hypergrid.py` as an example):

```
from gfn.env import DiscreteEnv
from gfn.states import States
from gfn.actions import Actions
import torch

VOCAB_SIZE = 5
BATCH_SIZE = 10
SEQ_LEN = 8


class MyEnv(DiscreteEnv):
    def __init__(self,**kwargs):
        super().__init__(**kwargs)

    def backward_step(self, states: States, actions: Actions) -> torch.Tensor:
        # Issue: need to find the iteration we're at to define backward step
        return states

    def update_masks(self, states: States) -> None:
        return torch.ones_like(states)

    def step(self, states: States, actions: Actions) -> torch.Tensor:
        return states


env = MyEnv(n_actions=VOCAB_SIZE,
            s0=torch.zeros(SEQ_LEN),
            state_shape=(SEQ_LEN,),
            action_shape=(1,),
            sf=torch.ones(SEQ_LEN) * torch.inf,
)
```

josephdviviano:
A few things: your environment does not need to be aware of the `BATCH_SHAPE` at all. The shape of an action is that of a single action, in this case, it's just `(1,)` -- selecting a token from your vocab. This is why in `env.py` both of these are true:

```
        assert dummy_action.shape == action_shape
        assert exit_action.shape == action_shape
```

[...]

PR #263, opened 2025-03-13
fixes edge case where submitted ints fail assert due to them not being tuples

josephdviviano:
Addresses the unexpected behaviour uncovered here:

https://github.com/GFNOrg/torchgfn/issues/228

### DIFF
--- a/src/gfn/env.py
+++ b/src/gfn/env.py
@@ -385,6 +385,13 @@ class DiscreteEnv(Env, ABC):
         if exit_action is None:
             exit_action = torch.tensor([n_actions - 1], device=device)
 
+        # If these shapes are integers, convert them to tuples.
+        if isinstance(action_shape, int):
+            action_shape = (action_shape,)
+
+        if isinstance(state_shape, int):
+            state_shape = (state_shape,)
+
         assert dummy_action is not None
         assert exit_action is not None
         assert s0.shape == state_shape


src/gfn/env.py
```diff
@@ -385,6 +385,13 @@ def __init__(
         if exit_action is None:
             exit_action = torch.tensor([n_actions - 1], device=device)
 
+        # If these shapes are integers, convert them to tuples.
+        if isinstance(action_shape, int):
+            action_shape = (action_shape,)
+
+        if isinstance(state_shape, int):
+            state_shape = (state_shape,)
+
```
hyeok9855, line 394: What about using assertion, i.e., `assert isinstance(action_shape, tuple)` ?

hyeok9855: Or we could add `int` to the type hints

saleml: I like @hyeok9855 's suggestion, of an assert. I do not think we should allow for non-tuple types here.

saleml (changes requested):

src/gfn/env.py
josephdviviano: I've added the type hints!

saleml: ok

saleml (approved):

PR #265, opened 2025-03-13
Refactoring w.r.t. `torch.device` in Env and Containers

hyeok9855:
I found an error due to incorrect device allocation. You can check by the following command in the `master` branch:

```bash
python tutorials/examples/train_hypergrid.py --loss TB
```

I’ve refactored the device allocation mainly in `Env` and `Container` (`Trajectories` and `Transitions`). Users are now required to match the device of the environments. 

The `env.device` is set by `s0.device`; this may be suboptimal, any suggestions are welcome.

### DIFF
--- a/src/gfn/env.py
+++ b/src/gfn/env.py
@@ -14,10 +14,6 @@ from gfn.utils.common import set_seed
 NonValidActionsError = type("NonValidActionsError", (ValueError,), {})
 
 
-def get_device(device_str, default_device) -> torch.device:
-    return torch.device(device_str) if device_str is not None else default_device
-
-
 class Env(ABC):
     """Base class for all environments. Environments require that individual states be represented as a unique tensor of
     arbitrary shape."""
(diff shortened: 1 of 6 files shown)


src/gfn/env.py
```diff
@@ -44,22 +39,20 @@ def __init__(
...
-            device_str: 'cpu' or 'cuda'. Defaults to None, in which case the device is
-                inferred from s0.
             preprocessor: a Preprocessor object that converts raw states to a tensor
                 that can be fed into a neural network. Defaults to None, in which case
                 the IdentityPreprocessor is used.
         """
-        self.device = get_device(device_str, default_device=s0.device)
-
-        self.s0 = s0.to(self.device)  # type: ignore
+        self.device = s0.device
```
younik, line 46: is `self.device` used anywhere now? Do we need it?

younik (approved):
Good catch!

src/gfn/env.py
hyeok9855: Yes. There are many usages of `env.device`, e.g., [here](https://github.com/GFNOrg/torchgfn/blob/f25409d4f6feaef7828751705bbf83a54015f53a/src/gfn/env.py#L459).

src/gfn/env.py
```diff
@@ -14,10 +14,6 @@
 NonValidActionsError = type("NonValidActionsError", (ValueError,), {})
 
 
-def get_device(device_str, default_device) -> torch.device:
```
saleml, line 17: this is clean!

saleml (approved):
Good catch. This removes redundancies!

PR #266, opened 2025-03-13
Refactor `__init__` of Transitions and Trajectories

hyeok9855:
These refactorings aim to make the behavior of `self._log_rewards` and `self.log_probs` clearer.

This branch contains the changes of #265 and shouldn't be merged before #265 being merged.

### DIFF
--- a/src/gfn/containers/trajectories.py
+++ b/src/gfn/containers/trajectories.py
@@ -10,7 +10,6 @@ from gfn.containers.state_pairs import StatePairs
 from gfn.containers.transitions import Transitions
 from gfn.env import Env
 from gfn.states import DiscreteStates, GraphStates, States
-from gfn.utils.common import has_log_probs
 
 
 # TODO: remove env from this class?
@@ -71,7 +70,6 @@ class Trajectories(Container):
         is used to compute the rewards, at each call of self.log_rewards
         """
         self.env = env
-        self.conditioning = conditioning
         self.is_backward = is_backward
 
         # Assert that all tensors are in the same device as the environment.
@@ -91,10 +89,17 @@ class Trajectories(Container):
             states if states is not None else env.states_from_batch_shape((0, 0))
         )
         assert len(self.states.batch_shape) == 2
+
+        self.conditioning = conditioning
+
         self.actions = (
             actions if actions is not None else env.actions_from_batch_shape((0, 0))
         )
-        assert len(self.actions.batch_shape) == 2
+        assert (self.actions.batch_shape == self.states.batch_shape == (0, 0)) or (
+            self.actions.batch_shape
+            == (self.states.batch_shape[0] - 1, self.states.batch_shape[1])
+        )
+
         self.when_is_done = (
             when_is_done
             if when_is_done is not None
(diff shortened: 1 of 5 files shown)


src/gfn/containers/trajectories.py
```diff
@@ -105,28 +110,35 @@ def __init__(
             and self.when_is_done.dtype == torch.long
         )
 
-        self._log_rewards = (
-            log_rewards
-            if log_rewards is not None
-            else torch.full(size=(0,), fill_value=0, dtype=torch.float, device=device)
-        )
-        assert (
+        # self._log_rewards can be torch.Tensor of shape (self.n_trajectories,) or None.
```
saleml, line 113: this is explicit. great!

src/gfn/containers/trajectories.py
```diff
@@ -184,15 +196,17 @@ def last_states(self) -> States:
     @property
     def log_rewards(self) -> torch.Tensor | None:
         """Returns the log rewards of the trajectories as a tensor of shape (n_trajectories,)."""
-        if self._log_rewards is not None:
-            assert self._log_rewards.shape == (self.n_trajectories,)
-            return self._log_rewards
-        if self.is_backward:
+        if self.is_backward:  # TODO: Why can't backward trajectories have log_rewards?
```
saleml, line 199: actually, you're right, there is nothing preventing us from doing it. Do you have a suggestion in mind for how to incorporate them to backward trajectories?

src/gfn/containers/trajectories.py
```diff
@@ -401,7 +413,7 @@ def to_transitions(self) -> Transitions:
             )
 
         # Initialize log_probs None if not available
-        if has_log_probs(self):
+        if self.log_probs is not None:
```
saleml, line 416: Are you sure about this change? `has_log_probs` checks if it not None and if it is not empty.

src/gfn/containers/transitions.py
```diff
@@ -182,6 +195,7 @@ def all_log_rewards(self) -> torch.Tensor:
         Raises:
             NotImplementedError: when used for backward transitions.
         """
+        # TODO: reuse self._log_rewards if it exists.
```
saleml, line 198: yes, good idea

src/gfn/containers/transitions.py
```diff
@@ -223,9 +237,7 @@ def __getitem__(
         is_done = self.is_done[index]
         next_states = self.next_states[index]
         log_rewards = self._log_rewards[index] if self._log_rewards is not None else None
-
-        # Only return logprobs if they exist.
-        log_probs = self.log_probs[index] if has_log_probs(self) else None
+        log_probs = self.log_probs[index] if self.log_probs is not None else None
```
saleml, line 240: same question here

saleml (commented):
Thanks for making these attributes more explicit and less confusing. A few questions remain.

src/gfn/containers/trajectories.py
hyeok9855: Not really, but we can imagine a trivial usecase: 
sample $(x,r) \sim \mathcal{B}$ (buffer)
sample $\tau \sim P_{B}(\tau|x)$

use the trajectory for training (via `Trajectories.reverse_backward_trajectories`)

src/gfn/containers/trajectories.py
hyeok9855: Let me check again.

hyeok9855: Yes, it seems that the changed version is correct.

When the `Trajectory` is empty, both `self.log_probs.shape` and `self.actions.is_dummy.shape` are [0, 0], so indexing is fine. The previous version using `has_log_probs` will set the `self.log_probs` to `None` in this case, which is not desirable.

When the `Trajectory` is not empty, `self.log_probs` should have same shape as `self.actions.batch_shape`.

src/gfn/containers/transitions.py
hyeok9855: See my answer above

younik (approved):

saleml (approved):
LGTM. Thanks

Issue #242, opened 2025-02-11
flow matching should allow for intermediate state reward calculation


josephdviviano:
flow matching loss calculation should accept a flag `calculate_intermediate_rewards=True`, which reflects the loss in equation 11 of https://arxiv.org/abs/2106.04399

PR #268, opened 2025-03-13
Replace `StatePairs` with `StatesContainer` for better indexing

hyeok9855:
### Key changes
~~`StatesWrapper`~~ `StatesContainer` accepts a single `states` tensor (in `StatePairs` accepts intermediary & terminating states, separately) and a boolean tensor `is_terminating` to indicate whether a state is terminating or not.

### Notes
~~Regarding the naming, I'm open to ideas (e.g., `StateContainer`?) or any other suggestions you might have!~~


src/gfn/containers/states_wrapper.py
```diff
@@ -0,0 +1,213 @@
...
+            else torch.zeros(
+                (len(self.states),), dtype=torch.bool, device=self.states.device
+            )
+        )
+        assert (
+            self.is_terminating.shape == self.states.batch_shape
+            and self.is_terminating.dtype == torch.bool
+        )
+
+        # self._log_rewards can be torch.Tensor of shape same as self.states or None.
```
josephdviviano, line 79: We can use self.has_log_rewards here

src/gfn/containers/states_wrapper.py
```diff
@@ -0,0 +1,213 @@
...
+            except NotImplementedError:
+                self._log_rewards[self.is_terminating] = torch.log(
+                    self.env.reward(self.last_states)
+                )
+
+        assert self._log_rewards.shape == self.states.batch_shape
+        return self._log_rewards
+
+    @property
+    def terminating_log_rewards(self) -> torch.Tensor:
```
josephdviviano, line 160: I suppose we are leaving open the ability to extend this to include non-terminating rewards, but this feature isn't implemented yet.

src/gfn/containers/trajectories.py
```diff
@@ -423,57 +440,59 @@ def to_transitions(self) -> Transitions:
...
             actions=actions,
-            is_done=is_done,
+            is_terminating=is_terminating,
             next_states=next_states,
             is_backward=self.is_backward,
             log_rewards=log_rewards,
             log_probs=log_probs,
         )
 
-    def to_states(self) -> States:
```
josephdviviano, line 433: wait - why did we remove this

src/gfn/containers/transitions.py
```diff
@@ -51,7 +51,7 @@ def __init__(
                 parents of the transitions.
             conditioning: The conditioning of the transitions for conditional MDPs.
             actions: Actions chosen at the parents of each transitions.
-            is_done: Tensor of shape (n_transitions,) indicating whether the action is the exit action.
+            is_terminating: Tensor of shape (n_transitions,) indicating whether the action is the exit action.
```
josephdviviano, line 54: why are we changing this terminology. `is_done` is used throughout the library.

src/gfn/containers/transitions.py
```diff
@@ -67,31 +67,35 @@ def __init__(
...
-            if is_done is not None
+        self.is_terminating = (
+            is_terminating
+            if is_terminating is not None
             else torch.full(size=(0,), fill_value=False, dtype=torch.bool, device=device)
         )
         assert (
-            self.is_done.shape == (self.n_transitions,)
-            and self.is_done.dtype == torch.bool
+            self.is_terminating.shape == (self.n_transitions,)
```
josephdviviano, line 97: I really would prefer we don't introduce changes in semantics like this. It's not required. Can you change it back please.

tutorials/examples/test_scripts.py
josephdviviano, line 1: These look like changes I introduced elsewhere.

josephdviviano (changes requested):
Really good PR! I like the implementation - very clean.

However I'm going to be annoying and ask that you undo all of the unnessicary re-naming of semantics. The class name `StatePairs` was good, and `is_done` is fine. We should try to avoid as much as possible renaming elements of the library like this, unless it generates genuine confusion into the codebase (in my opinion -- this isn't true here).

src/gfn/containers/__init__.py
```diff
@@ -1,13 +1,13 @@
 from .base import Container
 from .replay_buffer import NormBasedDiversePrioritizedReplayBuffer, ReplayBuffer
-from .state_pairs import StatePairs
+from .states_wrapper import StatesWrapper
```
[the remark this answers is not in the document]
hyeok9855: Thanks you for your review. I agree that `Wrapper` means nothing.

However, after these changes, I feel the name `StatePairs` might actually be misleading. Originally, `StatePairs` clearly described the class because it accepted two distinct sets of states during initialization. Now that the initialization only accepts a single states tensor and performs internal indexing, the name no longer clearly reflects the behavior.

@saleml @younik Do you have thoughts or suggestions on a more intuitive name?

src/gfn/containers/states_wrapper.py
hyeok9855: We don't have such method for now.
Also, we need to be careful in using `has_log_probs`; see [this discussion](https://github.com/GFNOrg/torchgfn/pull/266#discussion_r1993612096).

src/gfn/containers/states_wrapper.py
hyeok9855: Right, because we don't have a use case yet. Do you feel it's worth adding now, even without a clear use case?

src/gfn/containers/trajectories.py
hyeok9855: Since it isn't referenced anywhere. Would you prefer we keep it? why?

src/gfn/containers/transitions.py
hyeok9855: I wanted to use the same variable name for the indicator in both `Transitions` and `StatesWrapper`. **`Transitions.is_done` (before) and `StatesWrapper.is_terminating` represent the same concept**: both indicate whether a transition or state is ending. 

Specifically, in `Transitions`, `(s, a, s')` has `is_done=True` when `s'` is a sink state. Similarly, in `StatesWrapper`, the state `s` has `is_terminating=True` when its subsequent state is a sink. You can verify their equivalence by reviewing the `last_states` method in both classes.

***I'm open to standardizing on `is_done` if that's preferred, though I found `is_terminating` slightly clearer in this context. Let me know what you think!***

src/gfn/containers/transitions.py
```diff
@@ -67,31 +67,35 @@ def __init__(
...
             assert obj.tensor.device == device if obj is not None else True
-        for tensor in [conditioning, is_done, log_rewards, log_probs]:
+        for tensor in [conditioning, is_terminating, log_rewards, log_probs]:
             assert tensor.device == device if tensor is not None else True
 
         self.states = states if states is not None else env.states_from_batch_shape((0,))
         assert len(self.states.batch_shape) == 1
 
         self.conditioning = conditioning
+        assert (
```
saleml, line 81: does this have to be the case? I could easily think of scenarios where the condition is quite different from the states

src/gfn/containers/transitions.py
```diff
@@ -252,15 +268,24 @@ def __getitem__(
 
     def extend(self, other: Transitions) -> None:
         """Extend the Transitions object with another Transitions object."""
+        if self.conditioning is not None:
```
saleml, line 271: nice catch

src/gfn/containers/transitions.py
saleml: I don't have a strong opinion here. What about `when_is_done` in `Trajectories` ?

src/gfn/containers/trajectories.py
```diff
@@ -191,11 +195,15 @@ def max_length(self) -> int:
 
     @property
     def last_states(self) -> States:
+        """Get the last states, i.e. terminating states"""
```
saleml, line 198: Actually, I suggest to rename this function (even in the base class) to `terminating_states`. Consistency is good!

src/gfn/containers/trajectories.py
```diff
@@ -217,6 +225,9 @@ def __getitem__(
         when_is_done = self.when_is_done[index]
         new_max_length = when_is_done.max().item() if len(when_is_done) > 0 else 0
         states = self.states[:, index]
+        conditioning = (
```
saleml, line 228: good catch

src/gfn/containers/trajectories.py
```diff
@@ -378,8 +396,7 @@ def extend(self, other: Trajectories) -> None:
     def to_transitions(self) -> Transitions:
         """Returns a `Transitions` object from the trajectories."""
         if self.conditioning is not None:
-            traj_len = self.actions.batch_shape[0]
-            expand_dims = (traj_len,) + tuple(self.conditioning.shape)
+            expand_dims = (self.max_length,) + tuple(self.conditioning.shape)
```
saleml, line 399: This is better. thanks

src/gfn/containers/trajectories.py
saleml: I should have removed it when introducing states_pairs :)

src/gfn/containers/states_wrapper.py
saleml: @hyeok9855, please see https://github.com/GFNOrg/torchgfn/pull/262 (which should be merged before this one), where `has_log_rewards` exists

src/gfn/containers/states_wrapper.py
saleml: There is an inefficiency here: log_rewards are only useful for terminating states. Why should they be evaluated (on the fly) in `def log_rewards` for non-terminating states? 

What do you think of removing `terminating_log_rewards`, and having just `log_rewards`, with `-inf` value for non-terminating states?

saleml: It would also make sense for `to_states_wrapper` in `Trajectories` to have such a behavior of `log_rewards`: Each trajectory has a log_reward, that corresponds to that of its terminating state!

tutorials/examples/test_scripts.py
```diff
@@ -75,13 +75,21 @@ def test_hypergrid(ndim: int, height: int):
     args = HypergridArgs(ndim=ndim, height=height, n_trajectories=n_trajectories)
     final_l1_dist = train_hypergrid_main(args)
     if ndim == 2 and height == 8:
-        assert np.isclose(final_l1_dist, 8.78e-4, atol=1e-3)
+        assert np.isclose(
```
saleml, line 78: Thanks, this is useful

tutorials/examples/train_discreteebm.py
```diff
@@ -60,6 +60,7 @@ def main(args):  # noqa: C901
         preprocessor=env.preprocessor,
     )
     gflownet = FMGFlowNet(estimator)
+    gflownet = gflownet.to(env.device)
```
saleml, line 63: nice catch

saleml (commented):
Thank you for the PR. Please see my comments

src/gfn/containers/__init__.py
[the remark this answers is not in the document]
saleml: what about `StatesContainer` ?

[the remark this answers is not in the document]
younik: I also like StatesContainer!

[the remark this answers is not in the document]
hyeok9855: Now it's `StateContainer`

src/gfn/containers/transitions.py
hyeok9855: From today's meeting: use `is_terminating` and rename `when_is_done` to `terminating_idx`.

saleml:
Basically this: https://github.com/GFNOrg/torchgfn/issues/242 ?

From: Sanghyeok Choi ***@***.***>
Date: Friday, 14 March 2025 at 6:14 PM
To: GFNOrg/torchgfn ***@***.***>
Cc: Salem Lahlou ***@***.***>, Mention ***@***.***>
Subject: Re: [GFNOrg/torchgfn] Replace `StatePairs` with `StatesContainer` for better indexing (PR #268)

@hyeok9855 commented on this pull request.

________________________________

In src/gfn/containers/states_wrapper.py<https://cas5-0-urlprotect.trendmicro.com/wis/clicktime/v1/query?url=https%3a%2f%2fare01.safelinks.protection.outlook.com%2f%3furl%3dhttps%253A%252F%252Fgithub.com%252FGFNOrg%252Ftorchgfn%252Fpull%252F268%2523discussion_r1995657237%26data%3d05%257C02%257Csalem.lahlou%2540mbzuai.ac.ae%257C080c599b04e142ae183f08dd6302913c%257Cc93272d31b074b3da3b619b34a973915%257C0%257C0%257C638775584879839428%257CUnknown%257CTWFpbGZsb3d8eyJFbXB0eU1hcGkiOnRydWUsIlYiOiIwLjAuMDAwMCIsIlAiOiJXaW4zMiIsIkFOIjoiTWFpbCIsIldUIjoyfQ%253D%253D%257C0%257C%257C%257C%26sdata%3dLJ%252Fekb7fd5jjmUsMDJtTOEEtEFbvR%252B5xNVEOtP20eJg%253D%26reserved%3d0&umid=3bc2cf56-5a38-4669-ba71-b854f6d3af4e&rct=1741961696&auth=2558bcdb84e02b0c27cd7aa4822a24989cb4e596-3503d2f80dadfccf9dbf28ca161d5373269c189e>:

> +                device=self.states.device,

+            )

+            try:

+                self._log_rewards[self.is_terminating] = self.env.log_reward(

+                    self.last_states

+                )

+            except NotImplementedError:

+                self._log_rewards[self.is_terminating] = torch.log(

+                    self.env.reward(self.last_states)

+                )

+

+        assert self._log_rewards.shape == self.states.batch_shape

+        return self._log_rewards

+

+    @property

+    def terminating_log_rewards(self) -> torch.Tensor:

There is an inefficiency here: log_rewards are only useful for terminating states. Why should they be evaluated (on the fly) in def log_rewards for non-terminating states?

In this branch, the log_rewards function calls env.log_reward only for self.last_states, and only when self._log_rewards is None. So, the current implementation seems fine.

What do you think of removing terminating_log_rewards, and having just log_rewards, with -inf value for non-terminating states?

I'm wondering if there's a way to use the rewards from all the intermediate states as well. If that's possible, the current implementation seems clearer and more explicit.

src/gfn/containers/transitions.py
saleml: perfect

src/gfn/containers/transitions.py
hyeok9855: > does this have to be the case? I could easily think of scenarios where the condition is quite different from the states
This assertion ensures that each state has a distinct condition, which can vary across states.

However, I found this problematic because the dimensions of a state and a container can differ. I will fix this.

src/gfn/containers/trajectories.py
hyeok9855: Agree!

hyeok9855:
image

I didn't know that commit messages don't display things in `` ...

src/gfn/containers/transitions.py
hyeok9855: > However, I found this problematic because the dimensions of a state and a container can differ. I will fix this.

This is fixed

josephdviviano:
> Basically this: #242 ?

yes exactly. We can do it in a future PR though.

src/gfn/containers/states_wrapper.py
josephdviviano: Yes I agree, it will allow us to resolve #242

josephdviviano (approved):
lgtm!

Issue #249, opened 2025-03-03
Missing tests for `States`


josephdviviano:
We have tests for the `GraphStates` class but not `States` -- we should add a few (they will likely be very small due to our strict reliance on standard `torch.Tensor` ops).

Issue #198, opened 2024-10-11
Tests for replay buffers are not sufficient


josephdviviano:
A recent PR unconvered a fairly serious bug in the implementation of one replay buffer. I think we should have a close look at the associated tests, and write more to avoid this in the future.

Issue #112, opened 2023-08-02
Add tests for `States.__getitem__` and `States.__setitem__`


saleml:
As per title

hyeok9855:
Could you clarify what to test with `States` (and maybe `Actions` also)??
I'm asking because, for me, the `__getitem__` and `__setitem__` are basic tensor operations that don't seem to require tests.

saleml:
I suggested this because at some points, `__getitem__` with slices and boolean tensors was not working. I'm not sure it's needed anymore.
@josephdviviano, any opinion here?

josephdviviano:
I think it would make sense to have a set of tests which pass tensors of different types into `States` and `Actions` and ensure slicing ops return the right thing. I don't think it would need to be very sophisticated, runtime should be a few seconds.

Issue #108, opened 2023-08-02
Add a test for Trajectories extend function


saleml:
As was done in https://github.com/saleml/torchgfn/pull/79 for the `master` branch:

```
def test_extend_trajectories_on_cuda():
    import os
    import sys

    sys.path.insert(0, os.path.abspath("__file__" + "/../"))

    from src.gfn.containers.trajectories import Trajectories as Traj

    torch.manual_seed(0)

    env = HyperGrid(ndim=4, height=8, R0=0.01, device_str="cuda")
    sampler = TrajectoriesSampler(
        env=env,
        actions_sampler=DiscreteActionsSampler(
            estimator=LogitPFEstimator(env=env, module_name="NeuralNet"),
        ),
    )

    trajectories_1 = sampler.sample(n_trajectories=10)
    trajectories_2 = sampler.sample(n_trajectories=10)

    trajectories_1 = Traj(
        env=sampler.env,
        states=trajectories_1.states,
        actions=trajectories_1.actions,
        when_is_done=trajectories_1.when_is_done,
        is_backward=sampler.is_backward,
        log_rewards=trajectories_1.log_rewards,
        log_probs=trajectories_1.log_probs,
    )
    trajectories_2 = Traj(
        env=sampler.env,
        states=trajectories_2.states,
        actions=trajectories_2.actions,
        when_is_done=trajectories_2.when_is_done,
        is_backward=sampler.is_backward,
        log_rewards=trajectories_2.log_rewards,
        log_probs=trajectories_2.log_probs,
    )

    trajectories_1.extend(trajectories_2)
```

josephdviviano:
@younik can you look into whether there are missing tests here?

PR #271, opened 2025-03-14
Add tests for States, ReplayBuffer & Containers

younik:
Addresses https://github.com/GFNOrg/torchgfn/issues/249, https://github.com/GFNOrg/torchgfn/issues/198, https://github.com/GFNOrg/torchgfn/issues/112, https://github.com/GFNOrg/torchgfn/issues/108


josephdviviano:
I think the tests are failing due to a `pytest` version mismatch. I'm rewriting the tests to rely less on things like fixtures etc, which seem to be causing strange behaviour.

josephdviviano:
I'm less thinking it was due to a pytest version mismatch and rather that there was some conflicts between this branch and master, which I resolved. Note I had to change some of the tests which were not correct (they were getting caught by asserts we put into `master`). Therefore I'm going to ask that @saleml or @hyeok9855 double check this PR (I'm too close to it now to approve it in good conscience)

saleml (approved):
Lovely. Much needed PR. Thanks!

Issue #226, opened 2025-01-11
Add plotting and logging for hypergrid scripts and tutorials


saleml:
Seeing the training loop work is definitely a good sign, but is not enough. There could be many silent errors:
- the loss could explode
- even with a good loss, the learned distribution could be getting further from the target distribution

There is also no way to visualize the learned distributions, for the environment we use throughout our examples and tutorials, hypergrid.

I suggest we take inspiration from this [notebook](https://github.com/GFNOrg/GFN_vs_HVI/blob/master/grid/wandb_and_plots/download_data_wandb.ipynb) to implement a few utility functions, and end all hypergrid related scripts with plots.

What do you think about this @josephdviviano ?

josephdviviano:
I think this is a great idea and you did a great job on the PR :)

PR #276, opened 2025-03-15
Fix flow matching, add tests for hypergrid (FM and replay buffer), improve docs, validation/logging for hypergrid

saleml:
1- Flow Matching was broken. ~~I believe I have broken it when introducing StatePairs, later changed to StatesContainer.~~ This PR ~~doesn't solve it completely. BUT, it~~ adds some serious tests. We had FM silently broken before, and we need to prevent it in the future.

2- One problem I encountered while trying to fix FM is the following:
There is an issue with StatesContainer that makes FM not train
Consider 
```
trajectories =  Trajectories(n_trajectories=4, max_length=5
First 4 trajectories:
states=
[0 0]-> [0 1]-> [0 2]-> [1 2]-> [1 3]-> [-1 -1]
[0 0]-> [-1 -1]
[0 0]-> [-1 -1]
[0 0]-> [1 0]-> [-1 -1]
terminating_idx=[5 1 1 2])
```
We have `trajectories.log_rewards = tensor([-2.3026, -0.5108, -0.5108, -0.5108])`
Inside the call of `trajectories.to_states_container()`, 
```states.tensor = tensor([[0, 0],
        [0, 0],
        [0, 1],
        [1, 0],
        [0, 2],
        [1, 2],
        [1, 3]])
```
so this makes sense, because it's all non sink (dummy) states, and all non-initial states, except when they are terminating
but `log_rewards = tensor([-2.3026, -0.5108,    -inf, -0.5108,    -inf,    -inf, -0.5108])`

so this is wrong, because [1, 3] should be associated to the log_reward = -2.3026

So this PR fixes this issue

3- I add plotting to train_hypergrid. This include a nice little function that calculates the terminating state distribution P_T, that I added inside hypergrid.py

4- I add explanations for the three hypergrid scripts, and some sentences on them in README.


saleml:
Please note that this should address https://github.com/GFNOrg/torchgfn/issues/226

Also, naturally, the newly added tests will fail. So we need to fix FM before being able to merge this.

josephdviviano:
Hi @saleml - I don't fully understand the issue.

Can you point to a minimal example that will fail? I'd like to look into this.

I'm a bit confused because the `test_script.py` does successfully train flow matching models to the expected loss / `l1_dist`. But I would not be surprised that our test coverage is insufficient.

saleml:
Are you sure that test_script.py does successfully train flow matching models to the expected loss / l1_dist ?


Example:
`python train_hypergrid.py --loss FM --n_trajectories 64000 --ndim 2 --height 8 --plot`
final l1 loss: 0.009675217792391777

image

saleml:
Oh I see, it's the replay buffer!
`python train_hypergrid.py --loss FM --n_trajectories 64000 --ndim 2 --height 8 --plot --batch_size 16 --validation_interval 100 --replay_buffer_size 0` is ok:
image

saleml:
Actually, all good. No problem after all @josephdviviano , it's just that the default value of replay_buffer_size, 10, is ridiculously low.

All good, I can sleep soundly. I'll change the default value now

josephdviviano:
I'll review this soon -- thanks for the excellent PR!!

saleml:
> I'll review this soon -- thanks for the excellent PR!!

Thanks! This was my last commit for this PR (assuming all checks pass in github).

.github/workflows/ci.yml
```diff
@@ -32,4 +32,4 @@ jobs:
         run: pre-commit run --all-files
 
       - name: Test with pytest
-        run: pytest
\ No newline at end of file
+        run: pytest -x
```
josephdviviano, line 35: yeah good call

README.md
```diff
@@ -340,3 +340,65 @@ These methods are defined in `src/gfn/gflownet/base.py` and are abstract methods
 **Testing**
 
 Remember to create unit tests for your new GFlowNet to ensure it works as intended and integrates seamlessly with other parts of the codebase. This ensures maintainability and reliability of the code!
+
+
+## Training Examples
```
josephdviviano, line 345: again thanks for the much needed documentation

src/gfn/gflownet/flow_matching.py
```diff
@@ -84,6 +84,8 @@ def flow_matching_loss(
             AssertionError: If the batch shape is not linear.
             AssertionError: If any state is at $s_0$.
         """
+        if len(states) == 0:
+            return torch.tensor(0.0, device=states.device)
```
josephdviviano, line 88: nice - this resolves a bug I encountered last week actually. I think this is also something that can happen with all losses that rely on `Transitions` -- if the submitted batch has no valid transitions, the code cannot handle it (this is an edge case I mostly ran into when sampling an unlucky batch from a too-small replay buffer).

src/gfn/gflownet/flow_matching.py
```diff
@@ -164,6 +166,8 @@ def reward_matching_loss(
         log_rewards: torch.Tensor | None,
     ) -> torch.Tensor:
         """Calculates the reward matching loss from the terminating states."""
+        if len(terminating_states) == 0:
```
josephdviviano, line 169: ditto as above.

src/gfn/gym/hypergrid.py
```diff
@@ -199,6 +199,22 @@ def true_dist_pmf(self) -> torch.Tensor:
         true_dist /= true_dist.sum()
         return true_dist
 
+    def all_indices(self) -> List[Tuple[int, ...]]:
```
josephdviviano, line 202: this will likely need to get removed / merged with the similar functionality currently present in the `multinode` branch.

tutorials/examples/train_hypergrid.py
```diff
@@ -38,6 +43,70 @@
 DEFAULT_SEED = 4444
 
 
+def get_exact_P_T(env: HyperGrid, gflownet: GFlowNet) -> torch.Tensor:
```
josephdviviano, line 46: nice

josephdviviano (approved):
lgtm! I'm filing 2 small issues for future work.

Issue #189, opened 2024-10-08
Notesbooks out of sync with the torchgfn API


josephdviviano:
They should be fixed (any part that uses `torchgfn`).

younik:
This is solved with https://github.com/GFNOrg/torchgfn/pull/280

Issue #172, opened 2024-03-21
Fix `intro_gfn_continuous_line.ipynb`


saleml:
Running the second code cell returns `TypeError: Can't instantiate abstract class Line with abstract methods backward_step, step`

This is likely due to the recent changes in the environment API

josephdviviano:
Ah, so I also thought it was due to recent changes in the environment API, but it isn't -- it was never a functional notebook and is left as a todo for me.

I've added a note at the top of the notebook to this effect.

The goal of this notebook is to demonstrate how to handle continuous policies with more complex action spaces, e.g., sampling the likelihood of exiting at each step, sampling from a mixture beta distributions (instead of the other notebook where all trajectories are the same length, and the action space is just a single gaussian). The `_simple.ipynb` notebook is the Gaussian version (the one we presented at the workshop).

This notebook will be completed in a future PR, so I'll leave this open.

PR #280, opened 2025-03-17
Add CI for testing tutorial notebooks & fix notebooks

younik:
I am removing `intro_gfn_continuous_line.ipynb` as it is broken, and we don't plan to fix it before V2 release. This will solve https://github.com/GFNOrg/torchgfn/issues/172

Also related to https://github.com/GFNOrg/torchgfn/issues/272 and https://github.com/GFNOrg/torchgfn/issues/189

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,6 @@ wandb = { version = "*", optional = true }
 scikit-learn = {version = "*", optional = true }
 scipy = { version = "*", optional = true }
 matplotlib = { version = "*", optional = true }
-torchtyping = { version = "*", optional = true }
 
 [tool.poetry.extras]
 dev = [
@@ -69,7 +68,6 @@ dev = [
     "tox",
     "flake8",
     "nbmake",
-    "torchtyping"
 ]
 
 scripts = ["tqdm", "wandb", "scikit-learn", "scipy", "matplotlib"]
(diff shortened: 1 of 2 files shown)


josephdviviano:
can you move this notebook to a new branch?

pyproject.toml
```diff
@@ -43,13 +43,15 @@ sphinx_rtd_theme = { version = "*", optional = true }
...
 tox = { version = "*", optional = true }
+nbmake = { version = "*", optional = true } 
 
 # scripts dependencies.
 tqdm = { version = "*", optional = true }
 wandb = { version = "*", optional = true }
 scikit-learn = {version = "*", optional = true }
 scipy = { version = "*", optional = true }
 matplotlib = { version = "*", optional = true }
+torchtyping = { version = "*", optional = true }
```
josephdviviano, line 54: only comment - we should be able to fully remove this - if it's still in the notebooks, let's get rid of it.

josephdviviano (changes requested):
only ask is to remove `torchtyping`. looks good!

pyproject.toml
younik: good point, done!

hyeok9855 (approved):
LGTM. Thanks!

Issue #256, opened 2025-03-06
I think the `.batch_shape` should be equivalent to `.tensor.shape` or `.tensor.batch_shape` for graphs.


saleml:
I think the `.batch_shape` should be equivalent to `.tensor.shape` or `.tensor.batch_shape` for graphs.
If so, why do we need this setter for the `batch_shape`?
My suggestion is:
```python
@property
def batch_shape(self) -> tuple[int, ...]:
    return tuple(self.tensor.shape)[: -len(self.state_shape)]
```

_Originally posted by @hyeok9855 in https://github.com/GFNOrg/torchgfn/pull/252#discussion_r1983769864_

josephdviviano:
I'm also not sure about this one.

josephdviviano:
I disagree with my former self. I've implemented a fix that

- Does not permit the user to directly set the batch shape.
- Batch shape is mostly accessed from the underling tensor (in one case, we still use a `self._batch_shape` field.

PR #282, opened 2025-03-18
states/actions.batch_shape is set automatically

josephdviviano:
All of the elements where we explicitly set the `self.batch_shape` parameter are removed. This information is now gathered automatically and directly from the `self.tensor` field.

Addresses 
https://github.com/GFNOrg/torchgfn/issues/256

### DIFF
--- a/src/gfn/actions.py
+++ b/src/gfn/actions.py
@@ -314,10 +314,6 @@ class GraphActions(Actions):
         )
 
     def extend(self, other: Actions) -> None:
-        """Extends an Actions instance along the first dimension with dummy actions.
-
-        The Actions instance batch_shape must be 2-dimensional. This is used to pad
-        trajectories actions.
-        """
+        """Extends an Actions instance with another Actions instance."""
         super().extend(other)
         self._batch_shape = self.tensor.batch_size
--- a/src/gfn/actions.py
+++ b/src/gfn/actions.py
@@ -39,7 +39,10 @@ class Actions(ABC):
         ), f"Batched actions tensor has shape {tensor.shape}, but the expected action shape is {self.action_shape}."
 
         self.tensor = tensor
-        self.batch_shape = tuple(self.tensor.shape)[: -len(self.action_shape)]
+
+    @property
+    def batch_shape(self) -> tuple[int, ...]:
+        return tuple(self.tensor.shape)[: -len(self.action_shape)]
 
     @classmethod
     def make_dummy_actions(cls, batch_shape: tuple[int, ...]) -> Actions:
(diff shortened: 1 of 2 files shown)


src/gfn/actions.py
```diff
@@ -234,6 +232,10 @@ def __init__(self, tensor: TensorDict):
             batch_size=self.batch_shape,
         )
 
+    @property
+    def batch_shape(self) -> tuple[int, ...]:
+        return self._batch_shape
+
```
hyeok9855, line 238: Can't we use `self.tensor["action_type"].shape` ?

src/gfn/states.py
```diff
@@ -575,7 +559,7 @@ def __init__(self, tensor: GeometricBatch):
         """
         self.tensor = tensor
         if not hasattr(self.tensor, "batch_shape"):
-            self.tensor.batch_shape = self.tensor.batch_size
+            self.tensor.batch_shape = self.tensor.batch_size  # TODO: Is this correct?
```
hyeok9855, line 562: `torch_geometric.Batch.batch_size` is equivalent to the number of graphs, i.e., assuming 1d batch.

I think replacing this with an assertion `assert hasattr(self.tensor, "batch_shape")` will make things clearer.

hyeok9855 (approved):
Approve with minor comments. Thanks!

src/gfn/states.py
younik: `assert hasattr(self.tensor, "batch_shape")`  will fail most of the time

I think the point of this code is: if the user didn't specify a batch_shape, we assume a 1d batch.


I think the code should be changed like this tho:
```diff
- self.tensor.batch_shape = self.tensor.batch_size
+ self.tensor.batch_shape = (self.tensor.batch_size,)
```

younik (approved):
Same comments as @hyeok9855

src/gfn/actions.py
josephdviviano: I wasn't entirely sure if this would extend properly. Let me check.

josephdviviano: Hm, no this fails in many ways and I'm not really sure why. I don't mind this solution since there's no `@setter`. Feel free to investigate in this / a follow up PR.

src/gfn/states.py
josephdviviano: I've added 

```
        if not hasattr(self.tensor, "batch_shape"):
            if isinstance(self.tensor.batch_size, tuple):
                self.tensor.batch_shape = self.tensor.batch_size
            else:
                self.tensor.batch_shape = (self.tensor.batch_size,)

```

Issue #220, opened 2024-11-17
Preprocessor refactoring


younik:
I believe preprocessors should not be part of the Env class, and they should kept separate.

1) These lines in env examples don't scale well:
https://github.com/GFNOrg/torchgfn/blob/6f132a869902584c06b7f8054e76dda696d78c94/src/gfn/gym/hypergrid.py#L58-L72

2) It seems they are not used inside the env class, but they are just passed to GFNModule, which is a better location.

https://github.com/GFNOrg/torchgfn/blob/6f132a869902584c06b7f8054e76dda696d78c94/testing/test_gflownet.py#L40-L46

I believe they should be instantiated by the user and eventually passed to GFNModule.

josephdviviano:
Yes this makes sense to me. The GFNModule would be able to handle raw and
pre-processes states using whatever preprocessor is passed to it.

Joseph Viviano
@josephdviviano <https://twitter.com/josephdviviano>
viviano.ca

saleml:
By the way, this was the case 2 years ago. In fact, back then, we used to define an `Estimator` as a `GFModule` + `Preprocessor` object: https://github.com/GFNOrg/torchgfn/blob/7ac7de2a39e6bf6ca36985a3762fcef63d57bb56/gfn/estimators.py

[This big commit](https://github.com/GFNOrg/torchgfn/commit/c6d36517a640f0eb404cf89cd3310bab0df65721#diff-8afccb3dae31059eb5c5cdd762ad55eace9d95ae42f5c20e7e24bf28eff126b7), which is actually more of a code restructure than a mere commit, changed that. I had put the preprocessor in the environment.

Given the current structure of the code base, and the fact that `Estimator`s are not a thing anymore, I agree with with your suggestion.

PR #285, opened 2025-03-19
Remove preprocessor from environment

younik:
This PR removes the preprocessor from the env class, as suggested here: https://github.com/GFNOrg/torchgfn/issues/220

I am open to any feedback, i.e. even just rejecting this PR.

The drawback, is that it can be slightly harder for the user, see diff between before and now:
```diff
- env = HyperGrid(ndim=ndim, height=4, preprocessor_name='KHot')
+ env = HyperGrid(ndim=ndim, height=4)
+ preprocessor = KHotPreprocessor(env.height, env.ndim)
```

But, with these changes, it is easier to extend with a new preprocessor and clearer to the user what is happening.
Also, makes maintaining envs easier, as discussed in the issue.


josephdviviano:
I agree with this PR.

The workflow looks to me like (roughly):

```
env = Env(kw1=kw1, kw2=kw2, ..., kwn)
preprocessor = KHotPreprocessor(env.kw1, env.kw3)
module = Module()
estimator = Estimator(module=module, preprocessor=preprocessor)
gflownet = GFlowNet(estimator)
trajectories = gflownet.sample_trajectories(env, n)
```

josephdviviano:
However, I think you've introduced some bugs. The tests are not passing (maybe you didn't update all of the scripts quite properly).

josephdviviano (approved):
lgtm!

PR #288, opened 2025-03-21
Fix device for Graph

younik:
This is already on top of https://github.com/GFNOrg/torchgfn/pull/287
So merging this will update that branch @josephdviviano 

The current code doesn't work when setting cuda as device


josephdviviano (approved):
lgtm!

PR #291, opened 2025-03-24
Linear transformer module

josephdviviano:
Draft that adds a dependency and makes it easy for the user to include a linear transformer.

I think we can just work with this? Maybe we need a small tweak to have a controllable output dimension (a single linear layer after the call to self.module()). Let me know what you think!


josephdviviano:
@hyeok9855 you might want to use this in your branch now

saleml (approved):
LGTM

PR #296, opened 2025-03-26
Add DAG-GFlowNet (Bayesian Structure learning, Deleu et al., 2022)

hyeok9855:
### PR Summary

This PR includes the `BayesianStructure` env to reproduce [DAG-GFlowNet](https://arxiv.org/abs/2202.13903).

You can find a script for training with an MLP in `tutorials/examples/train_bayesian_structure.py`.

### Issues
- Training is very slow.

### TODO in some following PRs
- Make it run faster
- Use linear transformer policy
- Use ModifiedDB
- Reproduce [the results obtained with modified DB for larger graphs (#nodes=20 and 50)](https://arxiv.org/abs/2202.13903)

PR #299, opened 2025-04-02
Fix extend of GraphStates with 2d batch_shape

hyeok9855:
This fixes [this error](https://github.com/GFNOrg/torchgfn/pull/296#issuecomment-2761479074).

We shouldn't have simply concatenated two`data_list`s.

Here is an example:
Let's say, `max_len == 3`, `states1.batch_shape == (3, 2)`, and `states2.batch_shape == (3, 1)`.

Note that `GeometricData.to_data_list` should be aligned with the flattened index, which means that:
- `states1[0, 0] == states1.to_data_list[0]`
- `states1[0, 1] == states1.to_data_list[1]`
...
- `states1[2, 1] == states1.to_data_list[5]`

Similarly, for the concatenated states (`new_states`),
- `new_states[0, 0] == states1[0, 0] == new_states.to_data_list[0]`
- `new_states[0, 1] == states1[0, 1] == new_states.to_data_list[1]`
- `new_states[0, 2] == states2[0, 0] == new_states.to_data_list[2]`  <-- This was wrong.
...
- `new_states[2, 1] == states1[2, 1] == new_states.to_data_list[7]`  <-- This was wrong.
- `new_states[2, 2] == states2[2, 0] == new_states.to_data_list[8]`

### DIFF
--- a/src/gfn/states.py
+++ b/src/gfn/states.py
@@ -891,7 +891,7 @@ class GraphStates(States):
         else:
             # Handle the case where batch_shape is (T, B)
             # and we want to concatenate along the B dimension
-            assert len(self.batch_shape) == 2
+            assert len(self.batch_shape) == 2 and len(other.batch_shape) == 2
             max_len = max(self.batch_shape[0], other.batch_shape[0])
 
             # We need to extend both batches to the same length T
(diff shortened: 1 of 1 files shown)


younik (approved):
Aha, good catch!

You can also add the lines in the PR comments as tests, that would be nice to have

PR #300, opened 2025-04-03
synced changes from multinode branch ready for master

josephdviviano:
This adds a few major features:

- The Hypergrid environment is now compatible with multinode training.
- Losses can be handled in a multinode-friendly way.
- There are one or two small misc fixes that came up during the merge.

While development in the `multinode` branch is very much ongoing, the purpose of this PR is to sync the work done to-date with `master` to facilitate these otherwise quite painful merges.

I will be following up this PR with a smaller one which aims at reducing the length of time spent on `test_scripts.py` during CI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -178,6 +178,7 @@ cython_debug/
 
 # wandb
 wandb/
+wandb
 
 # slurm-specific stuff
 *.sh
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Never add this dependency the user might install manually here.
+torch-ccl/
+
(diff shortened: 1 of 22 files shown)


younik (approved):
Nothing seems off to me, but pre-commit is failing, so it must be fixed before merging.

I noticed you removed the ensure_same_device function in some places; this may lead to tests failures.

josephdviviano:
The tests are failing for an isort related reason.

josephdviviano:
fingers crossed this seems to be the issue (there was a `wandb/` folder at my root, uncommitted, that was created during debugging etc).

https://github.com/PyCQA/isort/issues/1497

PR #301, opened 2025-04-04
shortened the test_scripts.py run, and adjusted all targets appropria…

josephdviviano:
…tely

It what might be the most annoying PR I've ever done, I've made a pass at shortening the tests for CI. This required adjusting all target values to match. I've left the original targets in as comments for now, because I think we might find them useful if something is acting up in the future, but I can remove them if you want.

The tests run in a few minutes on my machine - let's see how long it takes on github.

### DIFF
--- a/tutorials/examples/test_scripts.py
+++ b/tutorials/examples/test_scripts.py
@@ -23,7 +23,7 @@ from .train_line import main as train_line_main
 
 @dataclass
 class CommonArgs:
-    batch_size: int = 16
+    batch_size: int = 128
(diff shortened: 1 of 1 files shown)


josephdviviano:
OK -- we're down from ~45 mins to ~13 mins. It's still long for my tastes but I think we should further improve this by removing whole iterations of a test, instead of tweaking `n_iterations` further, because the target numbers are already a bit flakey due to the models not being trained for long enough.

saleml (approved):
LGTM. Thanks for the change!

PR #302, opened 2025-04-08
Add assertions for GraphState.extend with 2d batch shape

hyeok9855:
This PR includes one missing commit that adds assertions for `GraphState.extend`, which is related to the issue that was resolved in #299.


younik (approved):
Thanks!

Issue #303, opened 2025-04-10
No reduction in GflowNets throw an error


Idriss-Malek:
Not reducing the loss (with mean or sum) throw an error:

> if torch.isnan(loss):
> RuntimeError: Boolean value of Tensor with more than one value is ambiguous

I guess it should be replaced by torch.isnan(loss).any()

saleml:
solved in #304

PR #304, opened 2025-04-10
Solves issue #303

Idriss-Malek:
Solves #303 
Recently, the option to not reduce, reduce by mean and reduce by sum the loss of TBGflowNet and DBGflowNet was added. 
In the case where no reduction was done, the `if torch.isnan(loss)` was throwing the error 

> RuntimeError: Boolean value of Tensor with more than one value is ambiguous

This PR solves this by adding `.any()` in the condition.


josephdviviano (approved):
LGTM ! Thanks

PR #306, opened 2025-04-16
Refactor graph-related codes

hyeok9855:
This PR:
1. resolves errors caused by incorrect device assignment when using GPUs.
2. includes several refactorings in `graph_building.py` and related areas throughout the repository.
3. enables the usage of epsilon-greedy (as well as temperature and sf-bias). 

Note that the performance of ring generation remains the same as before.

### DIFF
--- a/src/gfn/actions.py
+++ b/src/gfn/actions.py
@@ -41,25 +41,33 @@ class Actions(ABC):
         self.tensor = tensor
 
     @property
+    def device(self) -> torch.device:
+        return self.tensor.device
+
+    @property
     def batch_shape(self) -> tuple[int, ...]:
         return tuple(self.tensor.shape)[: -len(self.action_shape)]
 
     @classmethod
     def make_dummy_actions(
-        cls, batch_shape: tuple[int, ...], device: torch.device
+        cls, batch_shape: tuple[int, ...], device: torch.device | None = None
     ) -> Actions:
         """Creates an Actions object of dummy actions with the given batch shape."""
         action_ndim = len(cls.action_shape)
         tensor = cls.dummy_action.repeat(*batch_shape, *((1,) * action_ndim))
+        if device is not None:
+            tensor = tensor.to(device)
         return cls(tensor)
 
     @classmethod
     def make_exit_actions(
-        cls, batch_shape: tuple[int, ...], device: torch.device
+        cls, batch_shape: tuple[int, ...], device: torch.device | None = None
     ) -> Actions:
         """Creates an Actions object of exit actions with the given batch shape."""
         action_ndim = len(cls.action_shape)
         tensor = cls.exit_action.repeat(*batch_shape, *((1,) * action_ndim))
+        if device is not None:
+            tensor = tensor.to(device)
         return cls(tensor)
 
     def __len__(self) -> int:
@@ -68,10 +76,6 @@ class Actions(ABC):
     def __repr__(self):
         return f"""{self.__class__.__name__} object of batch shape {self.batch_shape}.
           The subclass did not implement __repr__."""
-
-    @property
-    def device(self) -> torch.device:
-        return self.tensor.device
 
     def __getitem__(
         self, index: int | slice | tuple | Sequence[int] | Sequence[bool] | torch.Tensor
@@ -211,6 +215,11 @@ class GraphActions(Actions):
             )
         self.tensor = tensor
 
+    @property
+    def batch_shape(self) -> tuple[int, ...]:
+        assert self.tensor.shape[-1] == 4
+        return self.tensor.shape[:-1]
+
     @classmethod
     def from_tensor_dict(cls, tensor_dict: TensorDict) -> GraphActions:
         """Creates a GraphActions object from a tensor dict."""
@@ -230,11 +239,6 @@ class GraphActions(Actions):
                 dim=-1,
             )
         )
-
-    @property
-    def batch_shape(self) -> tuple[int, ...]:
-        assert self.tensor.shape[-1] == 4
-        return self.tensor.shape[:-1]
 
     def __repr__(self):
         return f"""GraphAction object with {self.batch_shape} actions."""
--- a/src/gfn/actions.py
+++ b/src/gfn/actions.py
@@ -45,14 +45,18 @@ class Actions(ABC):
         return tuple(self.tensor.shape)[: -len(self.action_shape)]
 
     @classmethod
-    def make_dummy_actions(cls, batch_shape: tuple[int, ...]) -> Actions:
+    def make_dummy_actions(
+        cls, batch_shape: tuple[int, ...], device: torch.device
+    ) -> Actions:
         """Creates an Actions object of dummy actions with the given batch shape."""
         action_ndim = len(cls.action_shape)
         tensor = cls.dummy_action.repeat(*batch_shape, *((1,) * action_ndim))
         return cls(tensor)
 
     @classmethod
-    def make_exit_actions(cls, batch_shape: tuple[int, ...]) -> Actions:
+    def make_exit_actions(
+        cls, batch_shape: tuple[int, ...], device: torch.device
+    ) -> Actions:
         """Creates an Actions object of exit actions with the given batch shape."""
         action_ndim = len(cls.action_shape)
         tensor = cls.exit_action.repeat(*batch_shape, *((1,) * action_ndim))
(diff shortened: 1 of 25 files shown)


src/gfn/actions.py
```diff
@@ -45,14 +45,18 @@ def batch_shape(self) -> tuple[int, ...]:
         return tuple(self.tensor.shape)[: -len(self.action_shape)]
 
     @classmethod
-    def make_dummy_actions(cls, batch_shape: tuple[int, ...]) -> Actions:
+    def make_dummy_actions(
+        cls, batch_shape: tuple[int, ...], device: torch.device
```
josephdviviano, line 49: I feel a bit stupid reading this. How does the `device` actually get used?

src/gfn/actions.py
```diff
@@ -45,14 +45,18 @@ def batch_shape(self) -> tuple[int, ...]:
...
+    ) -> Actions:
         """Creates an Actions object of dummy actions with the given batch shape."""
         action_ndim = len(cls.action_shape)
         tensor = cls.dummy_action.repeat(*batch_shape, *((1,) * action_ndim))
         return cls(tensor)
 
     @classmethod
-    def make_exit_actions(cls, batch_shape: tuple[int, ...]) -> Actions:
+    def make_exit_actions(
+        cls, batch_shape: tuple[int, ...], device: torch.device
```
josephdviviano, line 58: Ditto.

src/gfn/actions.py
```diff
@@ -126,7 +130,9 @@ def extend_with_dummy_actions(self, required_first_dim: int) -> None:
             if self.batch_shape[0] >= required_first_dim:
                 return
             n = required_first_dim - self.batch_shape[0]
-            dummy_actions = self.__class__.make_dummy_actions((n, self.batch_shape[1]))
+            dummy_actions = self.__class__.make_dummy_actions(
+                (n, self.batch_shape[1]), device=self.device
```
josephdviviano, line 134: `self.device` is referencing the `self.tensor.device` - but AFACT the `tensor` does not have a device explicitly set anywhere.

src/gfn/actions.py
```diff
@@ -264,16 +270,20 @@ def edge_index(self) -> torch.Tensor:
...
 
     @classmethod
-    def make_dummy_actions(cls, batch_shape: tuple[int]) -> GraphActions:
+    def make_dummy_actions(
+        cls, batch_shape: tuple[int], device: torch.device
+    ) -> GraphActions:
         """Creates a GraphActions object of dummy actions with the given batch shape."""
         # TODO: make default dtype int32
-        tensor = torch.zeros(batch_shape + (4,), dtype=torch.int64)
+        tensor = torch.zeros(batch_shape + (4,), dtype=torch.int64, device=device)
```
josephdviviano, line 278: ok, you do it here -- is it just missing elsewhere?

src/gfn/gym/box.py
```diff
@@ -40,9 +40,12 @@ def __init__(
             exit_action=exit_action,
         )
 
-    def make_random_states_tensor(self, batch_shape: Tuple[int, ...]) -> torch.Tensor:
+    @staticmethod
+    def make_random_states_tensor(
+        batch_shape: Tuple[int, ...], device: torch.device
```
josephdviviano, line 45: can we draw this information from `self.device` (i.e., whatever device the env is initialized with) by default?

src/gfn/gym/discrete_ebm.py
```diff
@@ -108,14 +108,12 @@ def update_masks(self, states: DiscreteStates) -> None:
         states.backward_masks[..., : self.ndim] = states.tensor == 0
         states.backward_masks[..., self.ndim : 2 * self.ndim] = states.tensor == 1
 
-    def make_random_states_tensor(self, batch_shape: Tuple) -> torch.Tensor:
+    def make_random_states_tensor(
+        self, batch_shape: Tuple, device: torch.device
```
josephdviviano, line 112: can we draw this information from `self.device` (i.e., whatever device the env is initialized with) by default?

src/gfn/gym/graph_building.py
```diff
@@ -10,35 +11,7 @@
...
-        # Lower triangle.
-        i_lo, j_lo = torch.tril_indices(n_nodes, n_nodes, offset=-1, device=device)
-
-        ei0 = torch.cat([i_up, i_lo])  # Combine them
-        ei1 = torch.cat([j_up, j_lo])
-    else:
-        ei0, ei1 = torch.triu_indices(n_nodes, n_nodes, offset=1, device=device)
-
-    return ei0, ei1
+from gfn.utils.graphs import get_edge_indices
```
josephdviviano, line 14: cool

src/gfn/gym/graph_building.py
```diff
@@ -89,18 +62,6 @@ def __init__(
             is_directed=is_directed,
         )
 
-    def reset(
```
josephdviviano, line 92: Why remove this?

src/gfn/gym/hypergrid.py
```diff
@@ -131,17 +131,17 @@ def update_masks(self, states: DiscreteStates) -> None:
...
 
         Args:
             batch_shape: Tuple indicating the shape of the batch.
 
         Returns the batch of random states as tensor of shape (*batch_shape, *state_shape).
         """
-        return torch.randint(
-            0, self.height, batch_shape + self.s0.shape, device=self.device
-        )
+        return torch.randint(0, self.height, batch_shape + self.s0.shape, device=device)
```
josephdviviano, line 144: Same comment about it coming from the  env default device if the user does not pass the device.

src/gfn/preprocessors.py
```diff
@@ -31,9 +31,7 @@ def preprocess(self, states: States) -> torch.Tensor:
     def __call__(self, states: States | GraphStates) -> torch.Tensor | GeometricBatch:
         """Transform the states to the input of the neural network, calling the preprocess method."""
         out = self.preprocess(states)
-        if isinstance(out, GeometricBatch):
```
josephdviviano, line 34: Why is this being removed? There were cases where a GeometricBatch would be passed to this preprocessor by default. It seems reasonable to keep this as a default.

src/gfn/utils/distributions.py
```diff
@@ -49,51 +47,16 @@ def log_prob(self, sample: torch.Tensor) -> torch.Tensor:
 class GraphActionDistribution(Distribution):
     """A mixture distribution."""
 
-    def __init__(self, logits: TensorDict):
+    def __init__(self, probs: TensorDict):
         """Initializes the mixture distribution.
 
         Args:
-            logits: A TensorDict of logits.
+            probs: A TensorDict of probs.
```
josephdviviano, line 54: Where is the normalization coming from?

josephdviviano: Oh, I've seen you've moved the logic elsewhere. I'm curious why you would prefer the transformation of logits to probs there, as opposed to here (to me, this seems the logical spot, but I'm open to suggestions).

josephdviviano (approved):
LGTM! Thanks for the great work. Some comments, but I'm approving so I don't block the eventual merge.

src/gfn/actions.py
hyeok9855: Well, for `Actions` (the Tensor action), the device is automatically set the same as `cls.dummy_action` and `cls.exit_action`. We don't have such args for the `GraphActions`; that's why we need it as an argument.

I've changed the default to `None` for the `Actions`.

src/gfn/actions.py
hyeok9855: Here, we should ensure that `dummy_actions` and `self.tensor` are on the same device; thus, `self.device` will be fine. Am I misunderstanding your concern?

src/gfn/actions.py
hyeok9855: Well, I think I've made the changes needed.

src/gfn/gym/box.py
hyeok9855: I've also changed the default to `None` here. I'm open to any suggestions!

src/gfn/preprocessors.py
hyeok9855: Note first that the `Preprocessor.output_dim` should be equal to `GFNModule.module.input_dim`.

We can't simply assume that the graph action module always sets its `input_dim` to be equal to `out.x.shape[-1]`; **in `GraphEdgeActionMLP`, the `input_dim` should be `n_nodes**2`, not `1 == out.x.shape[-1]`**

src/gfn/actions.py
josephdviviano: where is the device from `tensor` set? Is it before we init this class instance?

src/gfn/gym/hypergrid.py
josephdviviano: ok thanks

josephdviviano:
thanks for taking the time to answer my questions. i'll merge since I made you wait so long.

src/gfn/actions.py
hyeok9855: Yes, it should be.

PR #308, opened 2025-04-17
Add batch_ptrs to GraphState

younik:


testing/test_states.py
```diff
@@ -480,6 +487,13 @@ def test_extend_2d(datas):
     assert (state1[2, 2].tensor.x == datas[8].x).all()
     assert (state1[2, 3].tensor.x == datas[9].x).all()
```
hyeok9855, line 488: Maybe add assertions like `(state1[2, 0].tensor.x == state1.sf.x).all()` and `(state1[2, 1].tensor.x == state1.sf.x).all()` ?

testing/test_states.py
```diff
@@ -480,6 +487,13 @@ def test_extend_2d(datas):
...
 
+    assert (state1[0, 0].tensor.edge_index == datas[0].edge_index).all()
+    assert (state1[0, 1].tensor.edge_index == datas[1].edge_index).all()
+    assert (state1[0, 2].tensor.edge_index == datas[4].edge_index).all()
+    assert (state1[0, 3].tensor.edge_index == datas[5].edge_index).all()
+    assert (state1[2, 2].tensor.edge_index == datas[8].edge_index).all()
+    assert (state1[2, 3].tensor.edge_index == datas[9].edge_index).all()
```
hyeok9855, line 495: Ditto

hyeok9855 (approved):
Impressive work! Thank you @younik.

I couldn't review every single line in `states.py`, but I've carefully checked the tests. We should check if this makes the replay buffer faster in #296 after merging this one.

younik:
> Impressive work! Thank you @younik.
> 
> I couldn't review every single line in `states.py`, but I've carefully checked the tests. We should check if this makes the replay buffer faster in #296 after merging this one.

Thanks! I addressed the comments.
I also checked with the ReplayBuffer, and it is around ~40% faster

PR #311, opened 2025-04-22
Fix init file in gym

alexandrelarouche:
I added PerfectBinaryTree and SetAddition to the `__init__.py` of `gym/` so they can be imported directly from `torchgfn.gym`


josephdviviano (approved):
lgtm!! thanks :)

PR #312, opened 2025-05-01
Added custom log_reward option to GFlowNets.

Idriss-Malek:
This pull request adds support for using custom rewards in GFlowNet implementations, enabling the use of intrinsic rewards for example (https://openreview.net/forum?id=HH4KWP8RP5).

## Changes

- Added a `log_rewards` argument to GFlowNet implementations.
  - By default, it is set to `None`, in which case the environment's reward is used. This preserves compatibility with existing code.
  - When provided, `log_rewards` is expected to be a tensor of the same shape as the container (e.g., `(n_trajectories,)`, `(n_transitions,)`, etc.), and it's used instead of the environment reward.

- In the DB and SubTB losses, I did not modify the forward-looking setup, except to rename the argument to `fl_log_rewards` to avoid naming conflicts.
  - I am not sure what this forward-looking setup does exactly (would appreciate any pointers to the original reference).
  - For now, I recommend using custom rewards with caution when using the forward-looking setup.

### DIFF
--- a/src/gfn/gflownet/base.py
+++ b/src/gfn/gflownet/base.py
@@ -1,6 +1,6 @@
 import math
 from abc import ABC, abstractmethod
-from typing import Any, Generic, Tuple, TypeVar
+from typing import Any, Generic, Optional, Tuple, TypeVar
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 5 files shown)


hyeok9855:
Hi @Idriss-Malek, thank you for submitting this PR!

Your current implementation adds arguments for the intrinsic rewards, which is clear, but I wonder if this is the best approach. What about creating a wrapper of env that enables adding the intrinsic reward to the original reward?

younik:
> > If we use 10 GFlowNets with 10 different intrinsic rewards, we will have 10 custom envs.
> 
> If we implement this as a wrapper that takes the original environment as an argument, then I think this would simply be 10 wrappers with a single custom environment?
> 
> Maybe we should gather more opinions to make a decision on this. @josephdviviano, @younik any thoughts?

I also favor the wrapper approach, assuming that the intrinsic reward depends only on the state of the environment, and not on the state of the net

Idriss-Malek:
> > > If we use 10 GFlowNets with 10 different intrinsic rewards, we will have 10 custom envs.
> > 
> > 
> > If we implement this as a wrapper that takes the original environment as an argument, then I think this would simply be 10 wrappers with a single custom environment?
> > Maybe we should gather more opinions to make a decision on this. @josephdviviano, @younik any thoughts?
> 
> I also favor the wrapper approach, assuming that the intrinsic reward depends only on the state of the environment, and not on the state of the net

The intrinsic reward doesn't always depend on the env. In https://openreview.net/forum?id=HH4KWP8RP5 for example, the intrinsic reward used to train a secondary gfn depends on the primary gfn.

Issue #315, opened 2025-05-14
Sampling with backward policy recovers log_rewards of initial_states


alexandrelarouche:
Hi,

I am currently playing around with backward sampling (i.e. sampling from the P_B) from some fixed terminating states. However, I encountered what I can only guess is a bug in the `Sampler` class implementation of `sample_trajectories`. It seems that currently, the log-rewards are computed for the initial state only, when the provided P_B estimator has the `is_backward` attribute set to True.

I am specifically referring to the following block: 
https://github.com/GFNOrg/torchgfn/blob/b9b62006c7ccaba540af0cead2f34795f759ffd3/src/gfn/samplers.py#L233-L244

The solve is straightforward, I think. We need to set the log_rewards at the states provided as input whenever we sample backward. This assumes that the provided states are terminating states (which seems reasonable, but maybe there are some use cases I am not aware of where this is not desirable).

Actionable items from this issue:
1. We need a test environment for backward sampling
2. When estimator.is_backward is True, we must not compute the log_reward at the end of the trajectory. Instead, we should set the `trajectories_log_rewards` to the rewards of the states that were provided as input to the `sample_trajectories` method.

I can implement both actionable items. However, I would like to clarify if the above assumption is a non-starter for the solution.

**Edit: Actually, I just noticed there is a catch if statement for backward trajectories saying they are not supported. So maybe this warrants a bigger issue.**
**Edit 2: Turns out, there is a `reverse_backward_trajectories` function available to translate backward trajectories to forward ones. This puts extra burden on the user, however.**

hyeok9855:
Hi, @alexandrelarouche. Thank you for raising this issue.

I agree that we can improve the backward sampling and how we handle the `log_rewards`. I would like to think about this more deeply, so I would appreciate it if you could share your specific use case involving backward sampling. 

Also, I want to hear your thoughts on our current approach; do you think we should allow learning directly from backward trajectories instead of converting them into forward ones using `reverse_backward_trajectories`?

alexandrelarouche:
Hey, thanks for being so open about discussing this.

### My use case
[...]

PR #316, opened 2025-05-15
Fix sampler rewards with backward estimator

alexandrelarouche:
Whenever `self.estimator.is_backward` was True, the log-rewards were computed with strictly the initial state of the environment (because it relied on the `new_dones` trick).

When entering `sample_trajectories` in `Sampler` we now check whether `estimator.is_backward` is True whenever states are provided. If it is, we set the log-rewards of the trajectories to be the reward of the provided `states`.

For this PR to work as intended, we need to assume that the provided `states` are in fact terminating states in the state-graph. (i.e. `states` $\subseteq \mathcal{X}$, since $R: \mathcal{X} \mapsto \mathbb{R}^+$).  Since not every environment implements the `terminating_states` property, we unfortunately cannot check that this is the case all the time.

If you have any thoughts on how to implement this check, or if you have improvements to suggests so as to make these changes better, your feedback is welcome.

Edit: Reference issue #315


src/gfn/samplers.py
```diff
@@ -144,6 +144,11 @@ def sample_trajectories(
                 len(states.batch_shape) == 1
             ), "States should have len(states.batch_shape) == 1, w/ no trajectory dim!"
             n_trajectories = states.batch_shape[0]
+            # Backward trajectories should have the reward at the beginning (terminating state)
+            if self.estimator.is_backward:
+                # [ASSUMPTION] all provided states are part of the terminating states (can be passed to reward fn)
+                # assert states in env.terminating_states # This assert would be useful, unfortunately, not every environment implements this.
+                trajectories_log_rewards = env.log_reward(states)
```
josephdviviano, line 151: I think this looks great!

josephdviviano (approved):
A good, simple PR. I'm sorry this wasn't merged before. I think this is a fairly parsimonious solution to the problem of backward sampling.

PR #317, opened 2025-05-16
Fix masking in SetAddition and PerfectBinaryTree envs

alexandrelarouche:
By testing with `reverse_backward_trajectories`, I noticed these environment's masking methods couldn't handle 2D batches of states (e.g. batches of states coming from a full, already realized trajectory).

I had to change the way masking was done in order to deal with the 2D boolean masking misbehaving with slicing.

I also added an option in `SetAddition` to have fixed trajectory lengths instead of allowing exit at any time step. I had already implemented it locally, but must have forgotten to add it to my last commit when I pushed these envs in the first place. Sorry!

All previous tests pass, and I verified that the current masks are the same as the previous implementations' masks.


josephdviviano (approved):
great stuff, thanks very much for the PR!

PR #318, opened 2025-05-20
Fix terminating states property for SetAddition environment

alexandrelarouche:
Sorry, in my last PR (#317), I forgot to add an `if` clause to the terminate_states property in `SetAddition`! My bad!


hyeok9855 (approved):
LGTM

hyeok9855:
Hi @alexandrelarouche, thank you for your contribution!

I think a more gentle introduction to the environment (and PerfectTree also) is essential, not only for users but also for effective debugging. Would you be able to create a tutorial script similar to `tutorials/examples/train_*.py`?

PR #320, opened 2025-05-23
Fix device checking in `Env` class

alexandrelarouche:
`ensure_same_device` used `sf` and `s0` as inputs, but only `self.sf` is sent to `s0`'s device


hyeok9855 (approved):
LGTM

PR #323, opened 2025-06-05
fix device mismatch in extend and stack of GraphStates

hyeok9855:

## Description
Fix the device mismatch that occurs when using CUDA in the `extend` and `stack` methods of `GraphStates`.

### DIFF
--- a/src/gfn/modules.py
+++ b/src/gfn/modules.py
@@ -524,7 +524,7 @@ class DiscreteGraphPolicyEstimator(GFNModule):
                 no_possible_edge_index, GraphActionType.ADD_EDGE
             ]
         ).all()
-        logits[GraphActions.EDGE_INDEX_KEY][no_possible_edge_index] = 0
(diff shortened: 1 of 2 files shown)


younik (commented):
Can't those operations happen on the GPU, if available?

hyeok9855:
> Can't those operations happen on the GPU, if available?

@younik, It seems `_inc_dict` is always on the CPU, even when the `Batch` is on the GPU (and thus it raises the device mismatch error when we use cuda). Could you double-check this for me?

hyeok9855:
@younik could you approve this PR?

younik:
I verified, and `_inc_dict` are created on the default device, which is CPU if not specified.

However, if you specify the CUDA to be default device `torch.set_default_device('cuda') ` this code will fail.
See: https://colab.research.google.com/drive/1Q5ilJccqA31y2arTX6emAbhiSdQ9_EEf#scrollTo=OpIrvuLSVOIF

Can we change the `.cpu()` code to be `.to(other.device)` instead?

younik (approved):
super, thanks!

PR #337, opened 2025-07-01
Use examples jdv

josephdviviano:
- [ ] I've read the [.github/CONTRIBUTING.md](.github/CONTRIBUTING.md) file
- [ ] My code follows the typing guidelines
- [ ] I've run pre-commit hooks locally

## Description
Have a hash-based method for checking mode coverage.

### DIFF
--- a/tutorials/examples/expert_data.py
+++ b/tutorials/examples/expert_data.py
@@ -8,9 +8,8 @@ import itertools
 import time
 from argparse import ArgumentParser, Namespace
 from collections import defaultdict
-from typing import Any, Dict, List, Literal, Tuple, Union
+from typing import Literal, Union
 
-import numpy as np
 import torch
 from tensordict import TensorDict
 from train_graph_ring import RingReward, init_env, init_gflownet, render_states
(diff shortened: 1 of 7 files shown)


tutorials/examples/expert_data.py
```diff
@@ -210,43 +259,58 @@ def main(args: Namespace):
     else:
         scheduler = None
 
+    # Generate all possible ring graphs.
+    ring_modes = generate_all_rings(args.n_nodes, device, args.replay_buffer_max_size)
+    n_modes = len(ring_modes)
+    mode_hashes = {graph_hash(m.tensor[0]) for m in ring_modes}  # Check mode coverage.
+    print(f"+ Number of modes: {n_modes}")
```
hyeok9855, line 266: Can't we compare the adjacency matrices instead of hashes?

josephdviviano:
We could they're just much more expensive to compute. I think it would be
sufficient for me to sort along the 1st dimension (for directed) and the
0th and 1st dimension independently (for undirected)

Joseph (Mobile)

PR #339, opened 2025-07-02
Improve docstrings

hyeok9855:

## Description
Improve docstrings, following the [Google style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).

### DIFF
--- a/src/gfn/containers/base.py
+++ b/src/gfn/containers/base.py
@@ -13,28 +13,58 @@ from gfn.env import Env
 
 
 class Container(ABC):
-    """Base class for states containers (states, transitions, or trajectories)."""
+    """Base class for state containers (states, transitions, or trajectories)."""
 
     @abstractmethod
     def __len__(self) -> int:
-        """Returns the number of elements in the container."""
+        """Returns the number of elements in the container.
+
+        Returns:
+            int: The number of elements in the container.
+        """
 
     @abstractmethod
     def __getitem__(
         self, index: int | slice | tuple | Sequence[int] | Sequence[bool] | torch.Tensor
     ) -> Container:
-        """Subsets the container."""
+        """Returns a subset of the container based on the provided index.
+
+        Args:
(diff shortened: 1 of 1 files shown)


hyeok9855:
FYI) I use the following prompt:

"""
Check the following Google-style Python docstrings:

```python
def function_with_pep484_type_annotations(param1: int, param2: str) -> bool:
    """Example function with PEP 484 type annotations.

    Args:
        param1: The first parameter.
        param2: The second parameter.

    Returns:
        The return value. True for success, False otherwise.
    """
    ...

class ExampleClass(object):
    """The summary line for a class docstring should fit on one line.

    If the class has public attributes, they may be documented here
    in an ``Attributes`` section and follow the same formatting as a
    function's ``Args`` section. Alternatively, attributes may be documented
    inline with the attribute's declaration (see __init__ method below).

    Properties created with the ``@property`` decorator should be documented
    in the property's getter method.

    Attributes:
        attr1 (str): Description of `attr1`.
        attr2 (:obj:`int`, optional): Description of `attr2`.

    """

    def __init__(self, param1: str, param2: int | None, param3: list[str]):
        """Example of docstring on the __init__ method.

        The __init__ method may be documented in either the class level
        docstring, or as a docstring on the __init__ method itself.

        Either form is acceptable, but the two should not be mixed. Choose one
        convention to document the __init__ method and be consistent with it.

        Note:
            Do not include the `self` parameter in the ``Args`` section.

        Args:
            param1 : Description of `param1`.
            param2 : Description of `param2`. Multiple
                lines are supported.
            param3 : Description of `param3`.

        """
        self.attr1 = param1
        self.attr2 = param2
        self.attr3 = param3  #: Doc comment *inline* with attribute

        #: list of str: Doc comment *before* attribute, with type specified
        self.attr4 = ['attr4']

        self.attr5 = None
        """str: Docstring *after* attribute, with type specified."""

```

Improve the docstrings in base.py. Please be clear and concise when you (re-)write the docstrings.
"""

younik (approved):
good to go for me

Issue #307, opened 2025-04-17
Training seems to have slowed down


sanchit-misra:
For the Feb 7 commit in multinode branch: #4b36589, if we run the following command, we would get to all 128 modes for the following command in 1800 iterations.

python -u train_hypergrid.py --ndim 7 --height 9 --R0 0.01 --tied --loss TB --n_trajectories 8192000 --batch_size 2048

But for the April 4 commit in multinode branch: #68d7429, for the same command, we get to 128 modes in ~2400 iterations.

Attaching the logs. You will notice extra stuff in logs. I have just added a few print statements but no other code changes. 

[apr-4-run.out.7.9.8192000.2048.64.txt](https://github.com/user-attachments/files/19789139/apr-4-run.out.7.9.8192000.2048.64.txt)

[feb-7-run.out.7.9.8192000.2048.16.txt](https://github.com/user-attachments/files/19789191/feb-7-run.out.7.9.8192000.2048.16.txt)

sanchit-misra:
Another interesting thing is that there is no run to run variation in the older commit #68d7429.

josephdviviano:
Interesting. And your measure of run to run variation is based on the loss,
or the validation metric only?

Joseph (Mobile)

sanchit-misra:
I traced the error back and found that even the sampled trajectories were different run to run after about 40 iterations. 
I think I found the reason for run to run variation. We need to add the following line in the set_seed method.

os.environ["PYTHONHASHSEED"] = str(seed)

It makes Python’s hashing deterministic, especially for operations involving hash-based data structures like dict and set. Once I added this line, I no longer see run to run variation for ndim = 6. I will try larger runs to confirm.

josephdviviano:
Thanks Sanchit this is a useful finding.

Note I have this already in the `improved_seeding` branch:
https://github.com/GFNOrg/torchgfn/blob/f89973a1715026d84a2bc34e71f7d99432b9d730/src/gfn/utils/common.py#L56

Can you give this implementation a shot and see if there is still an issue?
It fixes seeds for various libraries / types of hardware (no guarantees it
is complete though!)

Joseph Viviano
@josephdviviano <https://twitter.com/josephdviviano>
viviano.ca

josephdviviano:
Resolved in https://github.com/GFNOrg/torchgfn/pull/341

PR #341, opened 2025-07-03
added much more robust seed management utilities

josephdviviano:

## Description
Adds a much more robust set of seeding utilities that operates across various hardware configurations. This has resolved reproducibility issues we observed with large scale multi-node runs.


josephdviviano:
This fix was tested to fix run-to-run variation here https://github.com/GFNOrg/torchgfn/issues/307#issuecomment-3027954815

src/gfn/utils/common.py
```diff
@@ -1,23 +1,172 @@
...
 import random
+import threading
+import warnings
+from typing import Callable, Tuple
 
 import numpy as np
 import torch
+import torch.distributed as dist
+
+# -----------------------------------------------------------------------------
```
younik, line 11: the changes in common.py are helpful in a distributed settings. Should they go in a distributed module (just a file in utils called `distributed.py`)?

josephdviviano: I'm not sure I totally understand this one - the logic here is `if distributed: do distributed things, else: do non-distributed things` -- it means the user can just `set_seed(number)` and it will work regardless if the code is being run in the distributed or non-distributed context.

younik: I see, sounds good then

PR #355, opened 2025-07-24
Set dtype more easily

josephdviviano:

## Description

This PR is designed to allow users to easily change the precision of training library-wide. This can most easily be done by simply using `torch.set_default_dtype()` - or if the user wants to get more granular, the individual components of the library now accept a `dtype` field which set all floating point tensors to that dtype.

+ Removes any hard-coded float-based type. float dtypes are managed by `torch.set_default_dtype()`.
+ All int-based dtypes are consistently set to `long` - which is the pytorch default.
+ fixed various small documentation and linter bugs.
+ fixes a strange bug that showed up in `LocalSearchSampler` where `log_probs` would be `[]` instead of `None`.

I'm going to merge this into `multinode` first (where @chirayuharyan needs it) before getting the changes up to master. This contains @hyeok9855's recent PR.


josephdviviano:
gimmie a moment, I'm going to sync `multinode` and `master` which should clean up the diffs.

src/gfn/states.py
```diff
@@ -737,9 +744,13 @@ def tensor(self) -> GeometricBatch:
         """
         if self.data.size == 0:
             dummy_graph = GeometricData(
-                x=torch.zeros(0, self.num_node_classes, device=self.device),
+                x=torch.zeros(
```
josephdviviano, line 747: @younik @hyeok9855 I wasn't entirely sure if `x` and `edge_attr` should *always* be either `float` or `int`...

src/gfn/actions.py
```diff
@@ -425,8 +425,7 @@ def make_dummy_actions(
             A GraphActions object with the specified batch shape filled with dummy
             actions.
         """
-        # TODO: make default dtype int32
-        tensor = torch.zeros(batch_shape + (4,), dtype=torch.int64, device=device)
+        tensor = torch.zeros(batch_shape + (4,), dtype=torch.long, device=device)
```
younik, line 428: I think the TODO is important; int32 are faster than int64 on GPU

src/gfn/gym/bitSequence.py
```diff
@@ -284,7 +284,7 @@ def states_from_tensor(
         """
         if length is None:
             mask = tensor != -1
-            length = mask.int().sum(dim=-1)
+            length = mask.long().sum(dim=-1)
```
younik, line 287: this became long from int32(?)

src/gfn/gym/helpers/bayesian_structure/sampling.py
```diff
@@ -7,22 +7,22 @@
 def topological_sort_data(graph: GeometricData) -> list[int]:
     # Compute topological order for a torch_geometric graph DAG
     assert graph.node_names is not None, "Graph must have nodes"
-    indegree = torch.zeros(len(graph.node_names), dtype=torch.int)
+    in_degree = torch.zeros(len(graph.node_names), dtype=torch.long)
```
younik, line 10: same here, from int32 to long?

src/gfn/states.py
younik: neither, I believe, so this change looks good

src/gfn/utils/modules.py
```diff
@@ -422,7 +441,7 @@ def forward(self, states_tensor: GeometricBatch) -> TensorDict:
         node_features, batch_ptr = (states_tensor.x, states_tensor.ptr)
 
         # Multiple action type convolutions with residual connections.
-        x = self.embedding(node_features.squeeze().int())
+        x = self.embedding(node_features.squeeze().long())
```
younik, line 444: from int32 to long?

tutorials/examples/train_bayesian_structure.py
```diff
@@ -125,7 +125,7 @@ def forward(self, states_tensor: GeometricBatch) -> TensorDict:
         node_features, batch_ptr = (states_tensor.x, states_tensor.ptr)
 
         # Multiple action type convolutions with residual connections.
-        x = self.embedding(node_features.squeeze().int())
+        x = self.embedding(node_features.squeeze().long())
```
younik, line 128: ditto

tutorials/examples/train_bayesian_structure.py
```diff
@@ -309,7 +309,7 @@ def forward(self, states_tensor: GeometricBatch) -> TensorDict:
         )
         # batch_idx: (n_graphs * n_nodes)
 
-        node_embs = self.node_embedding(node_features.squeeze().int())
+        node_embs = self.node_embedding(node_features.squeeze().long())
```
younik, line 312: ditto

younik (approved):

src/gfn/actions.py
josephdviviano: What do you think of just putting everything to int32? I actually had it like this at one version of the PR - but I switched it because the default is int64 for pytorch in general so I was worried it was a potential footgun.

src/gfn/gym/bitSequence.py
josephdviviano: yes I moved everything to long (I want to unify the type - I'm fine with it being int32).

src/gfn/states.py
josephdviviano: ok actually, they're always a type of int now. we will need to add in the ability to make edge / node features continuous - issue filed here https://github.com/GFNOrg/torchgfn/issues/356

src/gfn/containers/replay_buffer.py
```diff
@@ -329,8 +331,12 @@ def add(self, training_objects: ContainerUnion):
                 training_objects = training_objects[idx_batch_batch]
 
                 # Compute all pairwise distances between the remaining batch & buffer.
-                batch = training_objects.terminating_states.tensor.float()
-                buffer = self.training_objects.terminating_states.tensor.float()
+                batch = training_objects.terminating_states.tensor.to(
+                    torch.get_default_dtype()
+                )
```
hyeok9855, line 336: It seems that we can remove these lines.

src/gfn/utils/prob_calculations.py
```diff
@@ -157,14 +157,17 @@ def get_trajectory_pfs(
...
+    trajectories: Trajectories,
+    fill_value: float = 0.0,
 ) -> torch.Tensor:
     """Calculates the log probabilities of backward trajectories.
 
     Args:
         pb: The backward policy estimator.
         trajectories: The trajectories to calculate probabilities for.
         fill_value: The value to fill for invalid states (e.g., sink states).
+        dtype: The dtype of the log probabilities.
```
hyeok9855, line 170: We should remove this.

src/gfn/utils/prob_calculations.py
```diff
@@ -300,7 +303,16 @@ def get_transition_pfs(
...
+def get_transition_pbs(
+    pb: Estimator, transitions: Transitions, dtype: torch.dtype | None = None
+) -> torch.Tensor:
+    """Calculates the log probabilities of backward transitions.
+
+    Args:
+        pb: The backward policy Estimator.
+        transitions: The transitions to calculate probabilities for.
+        dtype: The dtype of the log probabilities.
+    """
```
hyeok9855, line 315: The parameter dtype is not used; docstring should also be edited.

src/gfn/utils/training.py
```diff
@@ -13,13 +13,15 @@
...
-    env: DiscreteEnv, states: DiscreteStates
+    env: DiscreteEnv,
+    states: DiscreteStates,
 ) -> torch.Tensor:
     """Computes the empirical distribution of the terminating states.
 
     Args:
         env: The environment.
         states: The states to compute the distribution of.
+        dtype: The dtype of the PMF.
```
hyeok9855, line 24: Ditto

testing/test_containers.py
```diff
@@ -93,7 +93,8 @@ def trajectories_containers(env: Env):
                 [[1, 0], [0, 1]],  # Step 1
                 [[2, 0], [0, 2]],  # Step 2
                 [[2, 0], [0, 2]],  # Padding for trajectory 1
-            ]
+            ],
+            dtype=torch.int,  # Tests type compatibility with int32.
```
hyeok9855, line 97: Why do we want to test int32 when we only use the long?

src/gfn/samplers.py
```diff
@@ -28,12 +28,16 @@ class Sampler:
...
+    def __init__(
+        self,
+        estimator: Estimator,
+    ) -> None:
         """Initializes a Sampler with a PolicyEstimator.
 
         Args:
             estimator: The PolicyEstimator to use for sampling actions and computing
                 probability distributions.
+            dtype: The dtype of floating point numbers. If None, uses pytorch default fp dtype.
```
hyeok9855, line 40: We should remove this

src/gfn/states.py
```diff
@@ -710,14 +710,25 @@ class GraphStates(States):
...
+        data: np.ndarray,
+        categorical_node_features: bool = False,
+        categorical_edge_features: bool = False,
+        device: torch.device | None = None,
+    ) -> None:
         """Initializes the GraphStates with a numpy array of `GeometricData` objects.
 
         Args:
             data: A numpy array of `GeometricData` objects representing individual graphs.
+            dtype: The dtype of floating point numbers. If None, uses pytorch default dtype.
```
hyeok9855, line 724: Ditto

src/gfn/containers/replay_buffer.py
josephdviviano: i don't agree?

josephdviviano:
@hyeok9855 thanks for your comments!

src/gfn/containers/replay_buffer.py
hyeok9855: Because the `batch` is defined exactly the same above (line 316) and does not change, doesn't it?

src/gfn/utils/prob_calculations.py
hyeok9855: @josephdviviano, you missed this one!

PR #359, opened 2025-07-29
Multinode: average policies

younik:
First iteration of spawning new gflownets by averaging them


josephdviviano:
yikes this does look complicated.

FWIW i think we can get away with synchronous. looks like there's a simple approach using ray we could try:

```
import ray
import random
import torch
import numpy as np

@ray.remote
class PolicyNode:
    def __init__(self, policy, node_id):
        self.policy = policy
        self.node_id = node_id
    
    def get_weights(self):
        return {k: v.cpu().numpy() for k, v in self.policy.state_dict().items()}
    
    def set_weights(self, weights):
        state_dict = {k: torch.tensor(v) for k, v in weights.items()}
        self.policy.load_state_dict(state_dict)

def average_two_policies(policy_refs, new_policy):
    """Average weights from two randomly selected policies."""
    
    # Randomly select 2 policies
    selected_policies = random.sample(policy_refs, 2)
    
    # Get weights from selected policies
    weights_futures = [policy.get_weights.remote() for policy in selected_policies]
    weights_list = ray.get(weights_futures)
    
    # Average the weights
    averaged_weights = {}
    for key in weights_list[0].keys():
        averaged_weights[key] = (weights_list[0][key] + weights_list[1][key]) / 2.0
    
    # Set averaged weights to new policy
    new_policy.set_weights(averaged_weights)
    return new_policy

# Usage
def setup_distributed_policies():
    # Initialize your 4 policies on different nodes
    policy_refs = []
    for i in range(4):
        # Create policy instance (replace with your actual policy)
        policy = YourPolicyClass()  # Your policy implementation
        policy_ref = PolicyNode.remote(policy, f"node_{i}")
        policy_refs.append(policy_ref)
    
    # Create new policy
    new_policy = YourPolicyClass()
    new_policy_ref = PolicyNode.remote(new_policy, "new_averaged_policy")
    
    # Average weights from 2 random policies
    averaged_policy = average_two_policies(policy_refs, new_policy_ref)
    
    return averaged_policy
```

Let me know what you think

PR #381, opened 2025-09-02
Don't allow removal of nodes with edges

younik:
## Description
 

This PR implements the first solution described in https://github.com/GFNOrg/torchgfn/pull/380, i.e. it doesn't allow removal of graph nodes when there is an edge associated with it by proper action masking.

It actually still learn in 300 training steps, with hyperparam adjustment.


josephdviviano (approved):
thanks for the fix, looks great. will you add your experimental approach in a follow up PR?

Issue #190, opened 2024-10-09
conditioning example should do something


josephdviviano:
Right now, the conditioning example in `train_conditioning.py` is a red herring - the conditioning does nothing.

As suggested in a code review, we should make one of the conditions remove (at least one) of the modes in the hypergrid environment, so we can verify at test time that the conditioning is actually being used: https://github.com/GFNOrg/torchgfn/pull/188#discussion_r1788584163

This will require the conditioning to interact with the environment somehow which expands the scope of this a bit but should be interesting (and a high priority follow up to the original conditioning PR).

saleml:
or hypergrid with reward depending on x (a positive multiplier)

PR #385, opened 2025-09-05
Fix conditional GFlowNet training: implement functional continuous conditioning

saleml:
- Introduced new parameters in ConditionalArgs for learning rate, validation interval, and exploration.
- Updated test scripts to include various conditional training scenarios and validation checks.
- Enhanced train_conditional.py with a new ConditionalHyperGrid class for conditioning-aware rewards.
- Implemented validation during training to assess performance across different conditioning values.
- Added evaluation function for detailed metrics on conditional sampling distributions.

## Description
Addresses https://github.com/GFNOrg/torchgfn/issues/190
 ## Summary
  This PR fixes the non-functional conditioning in `train_conditional.py` and implements proper continuous conditioning
  that interpolates between uniform and multi-modal distributions.

  ## Problem
  The original conditioning implementation was decorative - conditioning values didn't actually affect the reward function,
   making the conditional GFlowNet behave identically regardless of conditioning.

  ## Changes

  ### Core Functionality
  - **Functional conditioning**: Conditioning now actually modifies rewards via linear interpolation
    - `conditioning=0`: Uniform distribution (all states get R0+R1+R2)
    - `conditioning=1`: Original HyperGrid multi-modal distribution
    - `conditioning∈(0,1)`: Linear interpolation between the two extremes

  ### Architecture Improvements
  - **LogZ depends only on conditioning**: Fixed LogZ to be a function of conditioning only (not states), matching the
  theoretical requirement for conditional GFlowNets
  - **Simplified validation**: Uses fresh trajectory samples for each conditioning value during validation
  - **Robust reward calculation**: Handles batch size mismatches that can occur during DetailedBalance loss calculation

  ### Training & Validation
  - Added hyperparameter arguments matching `train_hypergrid_simple.py`: `--lr`, `--lr_logz`, `--epsilon`,
  `--validation_samples`
  - Validation now samples fresh trajectories per conditioning value (doesn't reuse mixed-conditioning trajectories)
  - Concise logging: `Iter X: L1=[...] modes=Y` format
  - Mode discovery tracking for conditioning=1

  ### Testing
  - Updated test suite with functional tests (not just smoke tests)
  - Added tests for convergence, different dimensions, exploration, and all GFlowNet types
  - Fixed test compatibility issues with the `evaluate` flag

  ## Technical Details

  The key insight is that in conditional GFlowNets:
  - The reward function should vary based on conditioning
  - LogZ (partition function) should depend only on conditioning, not on states
  - During training, we sample conditioning uniformly from [0,1] for each batch

  ## Example Usage
  ```bash
  python train_conditional.py --ndim 2 --height 8 --epsilon 0.1 --lr 1e-3 --lr_logz 1e-2
```

  Testing

  All tests pass with the new implementation:
  - test_conditional_basic: Tests all GFlowNet types (TB, DB, SubTB, FM)
  - test_conditional_convergence: Verifies training converges to reasonable losses
  - test_conditional_different_dims: Tests with different environment dimensions
  - test_conditional_with_exploration: Tests with epsilon-greedy exploration
  - test_conditional_loss_types: Comprehensive test of all loss types

### DIFF
--- a/src/gfn/estimators.py
+++ b/src/gfn/estimators.py
@@ -191,6 +191,9 @@ class ScalarEstimator(Estimator):
         """
         return 1
 
+    def _calculate_module_output(self, input: States) -> torch.Tensor:
+        return self.module(self.preprocessor(input))
+
     def forward(self, input: States) -> torch.Tensor:
         """Forward pass of the module.
 
@@ -200,7 +203,7 @@ class ScalarEstimator(Estimator):
         Returns:
             The output of the module, as a tensor of shape (*batch_shape, 1).
         """
-        out = self.module(self.preprocessor(input))
+        out = self._calculate_module_output(input)
 
         # Ensures estimator outputs are always scalar.
         if out.shape[-1] != 1:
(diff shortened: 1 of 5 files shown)


saleml:
I don't see scenarios where logZ would depend on the state. In a non-conditional TB GFlowNets, logZ is simply a scalar, independent of the state. In a conditional one, it should be a function of the condition only. So perhaps we could make 'ConditionalLogZEstimator' as such and use it in this script ?

josephdviviano:
That said, the n_modes metric is 0 at the end of training. I'm not sure if that's just because the "mode found" logic is overly strict (it can be because there is aliasing of the grid, i.e., the true mode is between two grid locations and the reward value is never superthreshold). Are you seeing this on your side?

PR #388, opened 2025-09-15
Off policy methods

josephdviviano:

## Description

+ Adds `NoisyLayers` from `torchrl`.
+ Adds an example script comparing a few exploration methods.
+ This comparison script will optionally produce plots, or the following results table (with the default settings):

```
================================================================================
FINAL RESULTS SUMMARY
================================================================================
Experiment           L1 Dist         LogZ Diff       Loss            Modes          
Name                 (mean±std)      (mean±std)      (mean±std)      (mean±std)     
--------------------------------------------------------------------------------
epsilon_greedy       0.000±0.000    0.431±0.180    0.306±0.118    1.6±0.2
noisy_layers         0.000±0.000    0.175±0.037    0.401±0.193    3.5±0.1
on_policy            0.000±0.000    0.138±0.113    0.183±0.144    3.5±0.4
replay_buffer        0.000±0.000    0.204±0.082    0.280±0.178    3.5±0.5
================================================================================
BEST PERFORMERS:
  Lowest L1 Distance: epsilon_greedy
  Lowest LogZ Diff:   on_policy
  Lowest Loss:        on_policy
  Most Modes Found:   replay_buffer
================================================================================
```

### DIFF
--- a/src/gfn/utils/modules.py
+++ b/src/gfn/utils/modules.py
@@ -8,14 +8,28 @@ import torch.nn as nn
 from linear_attention_transformer import LinearAttentionTransformer
 from tensordict import TensorDict
 from torch_geometric.nn import DirGNNConv, GCNConv, GINConv
+from torchrl.modules import NoisyLinear
 
 from gfn.actions import GraphActions, GraphActionType
 from gfn.utils.common import is_int_dtype
 from gfn.utils.graphs import GeometricBatch, get_edge_indices
 
+ACTIVATION_FNS = {
+    "relu": nn.ReLU,
+    "leaky_relu": nn.LeakyReLU,
+    "tanh": nn.Tanh,
+    "elu": nn.ELU,
+}
+
 
 class MLP(nn.Module):
-    """Implements a basic MLP."""
+    """Implements a basic MLP with optional noisy layers for exploration.
+
+    When `trunk` is provided, the MLP will be a wrapper around the trunk.
+
+    See `Noisy Networks for Exploration (Fortunato et al., ICLR 2018)` for more details
+    on the noisy layers.
+    """
 
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


src/gfn/utils/modules.py
```diff
@@ -70,22 +102,34 @@ def __init__(
                     activation(),
                 ]
 
-            for _ in range(n_hidden_layers - 1):
-                arch.append(nn.Linear(hidden_dim, hidden_dim))
+            # Add the hidden layers. Put all noisy layers near the output.
+            n_noiseless_hidden_layers = n_hidden_layers - (n_noisy_layers - 1)
+            hidden_layer_types = [nn.Linear] * n_noiseless_hidden_layers + [
+                NoisyLinear
+            ] * (n_noisy_layers - 1)
```
hyeok9855, line 109: This will define the network differently from before if we set `n_noisy_layers`=1. We will add one more layer than we expected.

To resolve this, define `n_noisy_hidden_layers = max(0, n_noisy_layers - 1)` and then
```python
            n_noiseless_hidden_layers = (n_hidden_layers - 1) - n_noisy_hidden_layers
            hidden_layer_types = [nn.Linear] * n_noiseless_hidden_layers + [
                NoisyLinear
            ] * n_noisy_hidden_layers
```

hyeok9855 (changes requested):
Thanks @josephdviviano,
But please add your train_hypergrid_exploration_examples.py file 😊

src/gfn/utils/modules.py
```diff
@@ -34,30 +49,47 @@ def __init__(
...
+        # If a trunk is provided, the MLP will be a wrapper around the trunk.
+        if trunk is not None:
+            assert (
+                n_noisy_layers <= 1
+            ), "when trunk is provided, n_noisy_layers must be 0 or 1."
+            self.trunk = trunk
+            assert hasattr(trunk, "hidden_dim") and isinstance(
+                trunk.hidden_dim, torch.Tensor
+            ), "trunk must have a hidden_dim attribute"
+            self._hidden_dim = int(trunk.hidden_dim.item())
```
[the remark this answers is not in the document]
josephdviviano: Screenshot 2025-09-17 at 00 20 28

that would be nice but apparently we can't `__setattr__` on a `nn.Sequential` instance with a raw int.

src/gfn/utils/modules.py
josephdviviano: i think this line is wrong `n_noiseless_hidden_layers = (n_hidden_layers - 1) - n_noisy_hidden_layers` but the rest is correct, good catch.

hyeok9855 (approved):

src/gfn/utils/modules.py
josephdviviano: yes I had a bug before, but the `n_hidden_layers` count never included the `last_layer`. thanks!!

PR #392, opened 2025-09-17
Add TerminatingStateBuffer

hyeok9855:

## Description

Add `TerminatingStateBuffer`, which stores only the terminating states.

This could support potential use cases, such as:
1. When we have a dataset of terminating samples
2. When storing trajectories into the buffer requires too much memory
3. When we want to learn our sampler using multiple trajectories from a single terminating state

**Edit: Currently, learning with backward trajectory is sub-optimal in many aspects (e.g., backward sampling doesn't support `save_logprobs`, needs to call `reverse_backward_trajectories`, etc.). This point should be addressed in a subsequent PR.**

### DIFF
--- a/src/gfn/containers/replay_buffer.py
+++ b/src/gfn/containers/replay_buffer.py
@@ -388,7 +388,7 @@ class TerminatingStateBuffer(ReplayBuffer):
         conditioning = training_objects.conditioning
         log_rewards = training_objects.log_rewards
 
-        terminal_states_container = StatesContainer(
+        terminating_states_container = StatesContainer(
             env=self.env,
             states=terminating_states,
             conditioning=conditioning,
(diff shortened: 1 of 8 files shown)


src/gfn/containers/replay_buffer.py
```diff
@@ -143,6 +142,12 @@ def _add_objs(self, training_objects: ContainerUnion):
         assert self.training_objects is not None
         assert isinstance(training_objects, type(self.training_objects))  # type: ignore
 
+        # Clear fields that must be recomputed for Trajectories and Transitions.
```
josephdviviano, line 145: This should happen in the `TerminatingStateReplayBuffer` NOT the base class.

josephdviviano (changes requested):
LGTM except I don't want those fields to be cleared in the base ReplayBuffer class :) Should be specific to the class you added.

josephdviviano (approved):
lgtm!! thanks @hyeok9855 :)

PR #396, opened 2025-09-25
Add backward logprobs and estimator_outputs in Trajectories

hyeok9855:

## Description

1. Add `backward_logprobs` and `backward_estimator_outputs` in `Trajectories`, so that it doesn't have to be recomputed every time to get the loss.
2. We can input the precomputed `log_rewards` for backward sampling.
3. A few refactorings in Samplers
4. Remove `is_backward` flag from `Transitions` since we do not properly support it.

### TODO
Currently, we can store only one of the PF or PB during sampling (i.e., we always need to call our NN module for loss calculation). My next PR will resolve this by calculating both PF and PB within a single sampling process.

### DIFF
--- a/src/gfn/containers/trajectories.py
+++ b/src/gfn/containers/trajectories.py
@@ -471,6 +471,9 @@ class Trajectories(Container):
             A Transitions object with the same states, actions, and log_rewards as the
             current Trajectories.
         """
+        if self.is_backward:
+            return self.reverse_backward_trajectories().to_transitions()
+
         if self.conditioning is not None:
             # The conditioning tensor has shape (max_length, n_trajectories, 1)
             # The actions have shape (max_length, n_trajectories)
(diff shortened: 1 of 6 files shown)


src/gfn/containers/trajectories.py
```diff
@@ -142,13 +150,26 @@ def __init__(
...
             == self.actions.batch_shape
             and self.estimator_outputs.is_floating_point()
         )
 
+        self.backward_estimator_outputs = backward_estimator_outputs
+        assert self.backward_estimator_outputs is None or (
+            self.backward_estimator_outputs.shape[: len(self.states.batch_shape)]
+            == self.actions.batch_shape
+            and self.backward_estimator_outputs.is_floating_point()
+        )
```
josephdviviano: We could probably write a private `_check_estimator_outputs()` and `check_log_probs()` method which deduplicates this logic.

PR #398, opened 2025-09-27
sf dtype correctly set

josephdviviano:

## Description
sf now has dtype auto-set based on the submitted s0. also includes tests.

### DIFF
--- a/src/gfn/env.py
+++ b/src/gfn/env.py
@@ -6,7 +6,7 @@ from torch_geometric.data import Data as GeometricData
 
 from gfn.actions import Actions, GraphActions
 from gfn.states import DiscreteStates, GraphStates, States
-from gfn.utils.common import ensure_same_device, set_seed
+from gfn.utils.common import default_fill_value_for_dtype, ensure_same_device, set_seed
 
 # Errors
 NonValidActionsError = type("NonValidActionsError", (ValueError,), {})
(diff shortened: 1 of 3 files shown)


src/gfn/env.py
```diff
@@ -62,7 +62,10 @@ def __init__(
         self.s0 = s0
 
         if sf is None:
-            sf = torch.full(s0.shape, -float("inf"))
+            assert isinstance(s0, torch.Tensor), "When sf is None, s0 must be a Tensor"
+            sf = torch.full(
+                s0.shape, default_fill_value_for_dtype(s0.dtype), dtype=s0.dtype
+            )
```
younik, line 68: I am a bit scared about all the cases that don't support -inf, as it may result in silent errors.

For example, an env with integer dtype, we are using the min value, but this may be a valid state (maybe it is even the s0, so we are starting episodes that are already terminated). 

A possible alternative:
We just do `sf = torch.full(s0.shape, -float("inf"), dtype=s0.dtype)`, so we fail loudly when the dtype doesn't support the -inf. We can also try-catch to change the message prompting the user to specify the sf.

hyeok9855: > We just do sf = torch.full(s0.shape, -float("inf"), dtype=s0.dtype), so we fail loudly when the dtype doesn't support the -inf. We can also try-catch to change the message prompting the user to specify the sf.

I like this idea!

hyeok9855: A safe alternative would be just making `sf` a required argument for the base `Env` (and maybe for `DiscreteEnv` and `GraphEnv` as well).

PR #400, opened 2025-09-29
Generalize samplers

josephdviviano:

## Description
This factorizes the `Sampler` class to allow for custom `Adaptor` logic that changes how the sampler interfaces with the `Estimator` without requiring the user to subclass the entire `Sampler`. Here, a working example with `RecurrentAdaptor` is provided, which works with `RecurrentEstimators`. I've also added those, along with some example recurrent modules, and added some tests.

The following is taken from the default Adaptor class:
```
    The adapter bridges the generic sampling loop and the "classic" non‑recurrent
    estimators already used throughout the codebase. It exposes the minimal
    interface required by the `EstimatorAdapter` Protocol while keeping the
    sampler loop estimator-agnostic.

    Assumptions
    -----------
    - The wrapped estimator is non‑recurrent (no carry between steps).
    - If conditioning is provided, the estimator accepts `(states, conditioning)`;
      otherwise it accepts `(states)`.
    - The estimator provides `to_probability_distribution(states, est_out, **kw)`
      returning a torch Distribution over actions for the masked states.

    Context Lifecycle (opaque to the Sampler)
    ----------------------------------------
    The adapter owns an opaque context dict `ctx` which the Sampler never reads.
    The context is created once per rollout and mutated in place:

    - init_context(batch_size, device, conditioning) -> ctx
      Stores rollout invariants and optional conditioning. Also prepares per‑step
      buffers for artifacts that may be recorded (log_probs, estimator_outputs).

    - compute(states_active, ctx, step_mask, **policy_kwargs) -> (dist, ctx)
      1) Selects the appropriate estimator call signature depending on whether
         conditioning is present. If conditioning is present, the adapter slices
         it with `step_mask` so shapes match `states_active`.
      2) Calls the estimator forward pass to obtain the raw `est_out`.
      3) Converts `est_out` into a torch Distribution with
         `to_probability_distribution`.
      4) Saves `est_out` into `ctx["last_est_out"]` so it can optionally be
         recorded by `record_step` or exposed to legacy callers of `sample_actions`.

    - record_step(ctx, step_mask, sampled_actions, dist, save_logprobs, save_estimator_outputs)
      Materializes optional per‑step artifacts into ctx-managed buffers with
      mask‑aware padding back to the full rollout batch size `N`:
      * Log‑probs: computes `dist.log_prob(sampled_actions)` for active rows only,
        checks for `inf`, then writes into a 1D tensor of shape `(N,)` filled with
        zeros and masked assignment for active positions. Appends this to a list
        (one tensor per time step).
      * Estimator outputs: if requested, pads the last estimator output
        (`ctx["last_est_out"]`) to shape `(N, ...)` using `-inf` for inactive rows
        and appends to a list (one tensor per time step). This matches existing
        conventions elsewhere in the library.

    - finalize(ctx) -> {"log_probs": Tensor | None, "estimator_outputs": Tensor | None}
      Stacks recorded per‑step lists along the time dimension into tensors of shape
      `(T, N, ...)` suitable for `Trajectories`. Returns `None` for any artifact
      that was never recorded.

    Masking & Shapes
    ----------------
    - `states_active` always corresponds to `states[~dones]` inside the sampler.
    - The adapter receives `step_mask` (shape `(N,)`) to slice any step‑dependent
      inputs (e.g., conditioning) and to pad per‑step outputs to the full batch.
    - Padded tensors use `0.0` for log‑probs and `-inf` for estimator outputs to
      maintain compatibility with downstream code.

    Backward/Forward Direction
    --------------------------
    - `is_backward` is forwarded from the underlying estimator so the sampler can
      choose the appropriate environment transition (forward vs backward).

    Performance Notes
    -----------------
    - `ctx` is allocated once per rollout and mutated in place to avoid per‑step
      overhead.
    - If you know trajectory length bounds, you can extend this adapter to
      pre‑allocate fixed‑size storage in `init_context` rather than appending to
      Python lists.
```

NB: this merges changes from https://github.com/GFNOrg/torchgfn/pull/399, which should be merged first.

### DIFF
--- a/docs/source/guides/example.md
+++ b/docs/source/guides/example.md
@@ -22,14 +22,13 @@ env = HyperGrid(ndim=4, height=8)  # Grid of size 8x8x8x8
 preprocessor = KHotPreprocessor(ndim=env.ndim, height=env.height)
 
 # 2 - We define the needed modules (neural networks).
-input_dim = preprocessor.output_dim if preprocessor.output_dim is not None else env.state_shape[-1]
 module_PF = MLP(
-    input_dim=input_dim,
+    input_dim=preprocessor.output_dim,
     output_dim=env.n_actions
 )  # Neural network for the forward policy, with as many outputs as there are actions
 
 module_PB = MLP(
-    input_dim=input_dim,
+    input_dim=preprocessor.output_dim,
     output_dim=env.n_actions - 1,
     trunk=module_PF.trunk  # We share all the parameters of P_F and P_B, except for the last layer
 )
@@ -81,19 +80,18 @@ preprocessor = KHotPreprocessor(ndim=env.ndim, height=env.height)
 
 # 2 - We define the needed modules (neural networks).
 # The environment has a preprocessor attribute, which is used to preprocess the state before feeding it to the policy estimator
-input_dim = preprocessor.output_dim if preprocessor.output_dim is not None else env.state_shape[-1]
 module_PF = MLP(
-    input_dim=input_dim,
+    input_dim=preprocessor.output_dim,
     output_dim=env.n_actions
 )  # Neural network for the forward policy, with as many outputs as there are actions
 
 module_PB = MLP(
-    input_dim=input_dim,
+    input_dim=preprocessor.output_dim,
     output_dim=env.n_actions - 1,
     trunk=module_PF.trunk  # We share all the parameters of P_F and P_B, except for the last layer
 )
 module_logF = MLP(
-    input_dim=input_dim,
+    input_dim=preprocessor.output_dim,
     output_dim=1,  # Important for ScalarEstimators!
 )
 


src/gfn/estimators.py
```diff
@@ -798,3 +798,109 @@ def expected_output_dim(self) -> Optional[int]:
...
+        """Forward pass of the module.
+
+        Args:
+            states: The input states.
+            carry: The carry from the previous step.
+
+        Returns:
+            The output of the module, as a tensor of shape (*batch_shape, output_dim).
+        """
+        # TODO: Is this still true? NOTE: Can only be used for on-policy generation,
```
saleml, line 874: It looks like it. This forward is for on-policy generation only. For off-policy evaluation over entire trajectories, you need a separate path that handles padding/masks and batches multiple timesteps.

saleml:
Do you think it will be useful to add a small example in the docs on how to use a non-trivial adapter?

src/gfn/samplers.py
```diff
@@ -44,6 +407,7 @@ def sample_actions(
         conditioning: torch.Tensor | None = None,
```
younik, line 407: conditioning should be removed now as it goes in ctx?

src/gfn/samplers.py
```diff
@@ -77,36 +441,47 @@ def sample_actions(
...
-        actions = env.actions_from_tensor(actions)
+        # Allow adapter to record per-step artifacts for callers that reuse ctx.
+        self.adapter.record_step(
+            ctx=ctx,
+            step_mask=step_mask,
+            sampled_actions=actions_tensor,
+            dist=dist,
+            save_logprobs=save_logprobs,
+            save_estimator_outputs=save_estimator_outputs,
+        )
```
younik, line 474: why can't be done every `.compute` step?

younik (commented):
I’m not sure why this context is needed only during the sampling process and not within GFlowNets themselves. My concern is that we might be making a simplifying assumptio, such as only using on-policy sampling and not recomputing log probabilities, which is not clearly documented.

Also, I’m not an expert on GFlowNets, but does using an RNN here actually make sense? Does the math hold up?
Consider this scenario: two different trajectories lead to the same state s In that case, $g(s)$ would yield two different log probabilities depending on the trajectory, since the carry state differs.

I believe this PR was primarily motivated by the desire to support KV caching. That situation is somewhat different, because the assumption there is that the KV cache for a given state is always identical. For example, (1) in text generation, the state space is a tree (2) in molecule processing, nodes are visited in a fixed prefix order, so the order of adding nodes doesn’t change the resulting state.

I believe this could be handled easily by the user by adding a cache or carry attribute directly to the state object.

src/gfn/estimators.py
```diff
@@ -798,3 +798,109 @@ def expected_output_dim(self) -> Optional[int]:
...
+      adapter.
+    """
+
+    def __init__(
+        self,
+        module: nn.Module,
+        n_actions: int,
+        preprocessor: Preprocessor | None = IdentityPreprocessor(
+            output_dim=None
+        ),  # Addressed in https://github.com/GFNOrg/torchgfn/pull/399.
```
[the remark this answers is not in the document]
josephdviviano: This means `preprocessor` can be a `Preprocessor` or `None` in general, but in this case it's `IdentityPreprocessor(output_dim=None)`"

src/gfn/samplers.py
josephdviviano: they could maybe be merged, but I was thinking it was nice to have them in two different places (perhaps you're happy with the default record step logic but not with the default compute logic). the idea is that on average you shouldn't have to write too much code.

josephdviviano:
For the `kv` cache, it's not correct to put it in the state.

`RNN`-like policy may or may not make sense for your particular usecase (remember many people are particularly interested in tree-like MDPs where they definitely do make sense) - but this PR is also enabling *any* change in the interface between an estimator and the sampler, without requiring re-writing the whole sampler. I am already building on this, for example, to support the gfn chunking project.

josephdviviano:
"context is needed only during the sampling process and not within GFlowNets themselves"

This might be an issue!

younik (approved):
> `RNN`-like policy may or may not make sense

I am afraid of adding something plug-and-play that can be wrong silently, but I guess "we're all consenting adults" and we expect expert users.

josephdviviano:
TODO Look into using states or trajectories to contain the information in `ctx`, remove `ctx` completely, so that the logprob calcs do not depend on `ctx`.

src/gfn/estimators.py
josephdviviano: @saleml based on some conversations today - I'm going to look at changing the `ctx` design to use the trajectories container instead.

PR #402, opened 2025-10-01
hypergrid refactor

josephdviviano:

## Description
+ Moves key features involving distributed computing to `gfn.utils.distributed`.
+ Moves `Hypergrid` helper methods out of a dedicated folder into the main file.
+ Moves mode-counting logic to within the `Hypergrid` env, to allow for shared validation across all hypergrid scripts.
+ Moves standard `DiscreteEnvironment` evaluation code to the base class.
+ Adds strategy sampling to train_hypergrid.
+ Changes the training loop to support unique strategies per-agent, and agent weight restarting.

### DIFF
--- a/tutorials/examples/train_hypergrid.py
+++ b/tutorials/examples/train_hypergrid.py
@@ -32,13 +32,12 @@ import sys
 import time
 from argparse import ArgumentParser
 from math import ceil
-from typing import cast
+from typing import Tuple, cast
 
 import matplotlib.pyplot as plt
 import torch
 import torch.distributed as dist
 from matplotlib.gridspec import GridSpec
-from torch.nn.parallel import DistributedDataParallel as DDP
 from torch.profiler import ProfilerActivity, profile
 from tqdm import trange
 
@@ -62,7 +61,6 @@ from gfn.utils.distributed import (
     DistributedContext,
     gather_distributed_data,
     initialize_distributed_compute,
-    report_load_imbalance,
 )
 from gfn.utils.modules import MLP, DiscreteUniform, Tabular
 from tutorials.examples.multinode.hypergrid_diversity_score import (
@@ -72,12 +70,6 @@ from tutorials.examples.multinode.spawn_policy import (
     AsyncSelectiveAveragingPolicy,
     AverageAllPolicy,
 )
-
-
-def report_timing(all_timing_dict, world_size):
-    """Prints the timing information from the timing dictionary."""
-    # report_time_info(all_timing_dict, world_size)
-    report_load_imbalance(all_timing_dict, world_size)
 
 
 def get_exact_P_T(env: HyperGrid, gflownet: GFlowNet) -> torch.Tensor:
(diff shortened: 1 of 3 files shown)


src/gfn/gym/hypergrid.py
```diff
@@ -210,6 +225,106 @@ def reward(self, states: DiscreteStates) -> torch.Tensor:
         ), f"reward.shape is {reward.shape} and states.batch_shape is {states.batch_shape}"
         return reward
 
+    # -------------------------
+    # Mode utilities
+    # -------------------------
+    def _mode_reward_threshold(self) -> float:
```
younik, line 231: this method could be moved inside GridReward object and use polymorphism instead of the if chain

younik (approved):
I didn't review in detail, as it is hard for 1000+ LOC, assuming all of it comes from refactoring (only moving functions around).

I am not sure it is the case for sampling strategy, if you want a review for them consider moving it in another PR.

josephdviviano:
The sampling strategy PR should be merged first. It is a standalone PR

Joseph (Mobile)

src/gfn/gym/hypergrid.py
```diff
@@ -210,6 +225,106 @@ def reward(self, states: DiscreteStates) -> torch.Tensor:
...
+
+    def mode_mask(self, states: DiscreteStates) -> torch.Tensor:
+        """Boolean mask indicating which states are in a mode.
+
+        A state is flagged as mode if its reward is greater-or-equal to
+        the threshold based on `reward_fn_kwargs` (R0+R1+R2 by default).
+        """
+        rewards = self.reward(states)
+        threshold = self._mode_reward_threshold()
+        return rewards >= threshold
```
hyeok9855, line 295: We may want to subtract epsilon > 0, i.e., `rewards >= threshold - EPS`, to avoid floating point error. It happens sometimes (but I have no idea exactly when it does).

src/gfn/gym/hypergrid.py
```diff
@@ -210,6 +225,106 @@ def reward(self, states: DiscreteStates) -> torch.Tensor:
...
+            states.tensor.to(dtype=torch.get_default_dtype()) / (self.height - 1) > 0.5
+        ).long()
+        weights = (
+            2 ** torch.arange(self.ndim - 1, -1, -1, device=states.tensor.device)
+        ).long()
+        ids = (pos * weights).sum(dim=-1)
+        # Assign -1 to non-mode states
+        ids = torch.where(mask, ids, torch.full_like(ids, -1))
+
+        return ids
```
hyeok9855, line 315: I don't understand what's happening here. I guess this might only be valid for specific types of GridReward.

hyeok9855: Why do we need `mode_ids` or `modes_found`?

I guess the number of modes found can be obtained by `len(modes_visited)` where `modes_visited` is a set of states with rewards higher than a threshold.

hyeok9855 (commented):
I reviewed only env.py and gym, and left some comments. Thanks you!

tutorials/examples/train_hypergrid.py
```diff
@@ -687,48 +159,265 @@ def validate_hypergrid(
...
+    group = distributed_context.train_global_group
+    with torch.no_grad():
+        for t in _canonical_param_tensors_for_gflownet(gflownet):
+            buf = t.data.clone()
+            dist.all_reduce(buf, op=dist.ReduceOp.SUM, group=group)
+            buf -= t.data * G
+            buf /= N - G
+            t.data.copy_(buf)
+
+    return True
```
josephdviviano, line 355: To remove.

tutorials/examples/train_hypergrid.py
```diff
@@ -687,48 +159,265 @@ def validate_hypergrid(
...
+    assert gflownet is not None, f"gflownet is None, Args: {args}"
+    gflownet = gflownet.to(device)
+
+    # Choose initialization behavior: default to args.restart_init_mode, allow override.
+    mode = (
+        init_mode
+        if init_mode is not None
+        else getattr(args, "restart_init_mode", "random")
+    )
+    if mode == "mean_others":
```
josephdviviano, line 411: Needs to be replaced with Omar's work.

PR #419, opened 2025-10-17
Diffusion Sampling

hyeok9855:

## Description
Add diffusion sampling examples

## Todo
- Debug; for now, the learned sampler cannot approximate even a very easy 2d GMM target
- Add simple tests
- Add the Langevin Parameterization trick (which is a widely used technique in diffusion sampling tasks)
- Add evaluation metrics (ELBO, EUBO, Wasserstein distance)

## Todo maybe in follow-up PRs
- Test with prioritized terminal state replay buffer (e.g., Ch 3.3 of https://arxiv.org/abs/2510.11711)
- Learnable variance and Learnable PB (c.f., https://arxiv.org/abs/2506.01541)

### DIFF
--- a/tutorials/examples/test_scripts.py
+++ b/tutorials/examples/test_scripts.py
@@ -178,6 +178,9 @@ class DiffusionSamplerArgs:
     batch_size: int = 16
     lr: float = 1e-3
     lr_logz: float = 1e-1
+    eval_interval: int = 10
+    eval_n: int = 100
+    eval_batch_size: int = 100
     vis_interval: int = 10
     vis_n: int = 100
     visualize: bool = False


src/gfn/gym/diffusion_sampling.py
```diff
@@ -0,0 +1,388 @@
...
+        target: The target distribution.
+        device: The device to use.
+        check_action_validity: Whether to check the action validity.
+    """
+
+    def __init__(
+        self,
+        target_str: str,
+        target_kwargs: dict[str, Any],
+        device: torch.device = torch.device("cpu"),
```
younik: I think usually the default device should be the one of torch, not cpu

src/gfn/gym/diffusion_sampling.py
```diff
@@ -0,0 +1,388 @@
...
+        """Check if the actions are valid.
+
+        Args:
+            states: The current states (not used here).
+            actions: The actions to check.
+
+        Returns:
+            True if the actions are valid, False otherwise.
+        """
+        assert len(states.batch_shape) == 1, "States must have a batch_shape of length 1"
```
younik: why?

src/gfn/gym/diffusion_sampling.py
hyeok9855: Why? Other envs (like Box, HyperGrid, ...) are also using cpu as the default.

src/gfn/gym/diffusion_sampling.py
```diff
@@ -0,0 +1,388 @@
...
+        """
+        batched = x.ndim == 2
+        if not batched:
+            x = x.unsqueeze(0)
+
+        log_probs = self.distribution.log_prob(x)
+        if not batched:
+            log_probs = log_probs.squeeze(0)
+
+        return log_probs
```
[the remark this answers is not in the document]
hyeok9855: Oh, I omitted the log_reward from the Env. I'm fixing it now!

tutorials/examples/train_diffusion_sampler.py
```diff
@@ -0,0 +1,536 @@
...
+    def __init__(
+        self,
+        s_dim: int,  # dimension of states (== target.dim)
+        harmonics_dim: int = 64,
+        t_emb_dim: int = 64,
+        s_emb_dim: int = 64,
+        hidden_dim: int = 64,
+        joint_layers: int = 2,
+        zero_init: bool = False,
+        # predict_flow: bool,  # TODO: support predict flow for db or subtb
```
younik: Just to bring to attention there are a bunch of TODO here
It is fine to merge them to master

tutorials/examples/train_diffusion_sampler.py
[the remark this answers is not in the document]
josephdviviano: yes I think we need a test or two.

tutorials/examples/train_diffusion_sampler.py
```diff
@@ -0,0 +1,517 @@
...
+    def forward(self, preprocessed_states: torch.Tensor) -> torch.Tensor:
+        return torch.zeros_like(preprocessed_states[..., :-1])
+
+
+###########################################
+### Distribution Wrapper and Estimators ###
+###########################################
+
+
+class IsotropicGaussianWithTime(Distribution):
```
josephdviviano: As this (and the brownian stuff below) should be added to the main library for sure, since as I understand it, it will be generally useful for diffusion modelling.

josephdviviano (changes requested):
Small changes requested (happy to help with this as well). Great work!

src/gfn/gym/diffusion_sampling.py
hyeok9855: What about not setting the default value and forcing the user to specify it?

src/gfn/gym/diffusion_sampling.py
```diff
@@ -0,0 +1,388 @@
...
+
+        Args:
+            batch_size: The number of samples to sample.
+            seed: The seed for the random number generator.
+
+        Returns:
+            The samples.
+        """
+        if seed is not None:
+            set_seed(seed)
```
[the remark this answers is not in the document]
hyeok9855: I hope we have a context manager that allows us to use an independent random seed that is used only inside a "with ~" block but has no effect outside of it. I will leave a comment.

tutorials/examples/train_diffusion_sampler.py
```diff
@@ -0,0 +1,536 @@
...
+        super().__init__()
+        self.input_dim = s_dim + 1  # + 1 for time, for the default IdentityPreprocessor
+
+    def forward(self, preprocessed_states: torch.Tensor) -> torch.Tensor:
+        return torch.zeros_like(preprocessed_states[..., :-1])
+
+
+###########################################
+### Distribution Wrapper and Estimators ###
+###########################################
```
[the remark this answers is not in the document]
hyeok9855: same as above

tutorials/examples/train_diffusion_sampler.py
josephdviviano: See my above comment. What do you think? I feel like it's likely we will end up adding more environments longer term that will benefit from these modules, so might as well centralize now.

src/gfn/gym/diffusion_sampling.py
josephdviviano: I think I agree with @younik here. We should have all of these things default to the same device, which should be a hardware accelerator if it exists.

josephdviviano (approved):
lgtm @hyeok9855 - I've left some more comments for you but I'll approve now to unblock you. merge when you think it's ready :)

hyeok9855:
I made a bunch of refactorings.

I also found that our implementation learns quite slowly (many more epochs are needed to reach a certain level of loss) compared to the original implementation when tested with the same learning rates, while they behave similarly in the long run (mode-collapse, for the 32d ManyWell target).

I'll keep this issue in mind and look into it more in the follow-up PRs.

josephdviviano:
> I also found that our implementation learns quite slowly (many more epochs are needed to reach a certain level of loss) compared to the original implementation when tested with the same learning rates

Are all other elements matched? (Architecture, optimizer, noise parameters).

hyeok9855:
Let me correct myself; it is slightly slower. But this is based only on my visual inspection (using visualisation of the first two dimensions). I will investigate again after adding some evaluation metrics (will do this week).

tutorials/examples/train_diffusion_sampler.py
hyeok9855: See my comments above!

tutorials/examples/train_diffusion_sampler.py
```diff
@@ -0,0 +1,536 @@
...
+from gfn.samplers import Sampler
+from gfn.states import States
+from gfn.utils.common import set_seed
+
+LOGTWOPI: float = math.log(2 * math.pi)
+
+
+##############################
+### Neural Network Modules ###
+##############################
```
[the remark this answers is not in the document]
josephdviviano: I'd be interested to help build these extensions once I'm satisfied with the multinode / multiagent stuff :)

josephdviviano:
> For example, at epoch 1999, our implementation gives a sampler like this: image
> 
> and the base implementation gives: image
> 
> Our implementation shows a similar mode-collapse behavior after >= 4000 epochs.

I've raised an issue, we can debug in a follow up PR.

Issue #424, opened 2025-10-24
Device errors in testing on a GPU environment.


hyeok9855:
I tested this on the Mila cluster (node: cn-l014) with a l40s GPU.
Check the log below for details.

---
```
(torchgfn) chois@cn-l014:~/workspace/torchgfn$ pytest testing/
====================================================================================== test session starts ======================================================================================
platform linux -- Python 3.11.11, pytest-8.4.1, pluggy-1.6.0
rootdir: /home/mila/c/chois/workspace/torchgfn
configfile: pyproject.toml
plugins: nbmake-1.5.5
collected 1856 items                                                                                                                                                                            

testing/test_actions.py ..                                                                                                                                                                [  0%]
testing/test_adaptor_estimator_gflownet_integration.py ......                                                                                                                             [  0%]
testing/test_containers.py ...                                                                                                                                                            [  0%]
testing/test_environments.py .....................................                                                                                                                        [  2%]
testing/test_estimators.py ........................................                                                                                                                       [  4%]
testing/test_gflownet.py ...                                                                                                                                                              [  4%]
testing/test_modules.py ..FF..FF                                                                                                                                                          [  5%]
testing/test_parametrizations_and_losses.py ...........s.s...s.s......................................................................................................................... [ 12%]
......................................................................................................................................................................................... [ 22%]
..................................................ssss........................................ssss........................................ssss........................................sss [ 32%]
s........................................ssss........................................ssss........................................ssss........................................ssss........ [ 42%]
...............................sssss...s...s...s...s...s...s...s...s...s...sssss...s...s...s...s...s...s...s...s...s...sssss...s...s...s...s...s...s...s...s...s...sssss...s...s...s...s. [ 52%]
..s...s...s...s...s...sssss...s...s...s...s...s...s...s...s...s...sssss...s...s...s...s...s...s...s...s...s...sssss...s...s...s...s...s...s...s...s...s...sssss...s...s...s...s...s...s.. [ 62%]
.s...s...s...................................................................................................s...s...s...s...s...s...s...s...s...s...s...s....                            [ 71%]
testing/test_probability_calculations.py .........                                                                                                                                        [ 71%]
testing/test_replay_buffer.py .......s.....                                                                                                                                               [ 72%]
[...]

PR #426, opened 2025-10-30
Device fix

josephdviviano:

## Description
+ device handling is now done consistently across states classes.
+ added new tests to ensure device handling works properly with cuda.
+ fixes all failing tests (addressing https://github.com/GFNOrg/torchgfn/issues/424)

### DIFF
--- a/src/gfn/states.py
+++ b/src/gfn/states.py
@@ -755,22 +755,60 @@ class GraphStates(States):
         self.categorical_node_features = categorical_node_features
         self.categorical_edge_features = categorical_edge_features
         self.data = data
+
         # Resolve device per instance: prefer explicit, else infer, else default
-        inferred: torch.device | None = None
+
+        if device is not None:
+            resolved_device = device
+        else:
+            inferred_device: torch.device | None = None
+
+            # Get the device from the first graph in the data array.
+            if data.size > 0:
+                g = data.flat[0]
+                assert isinstance(g, GeometricData)
+                assert isinstance(g.x, torch.Tensor)
+                inferred_device = g.x.device
+
+            if inferred_device is not None:
+                resolved_device = inferred_device
+            else:
+                resolved_device = torch.empty(0).device
+
+        self._device = resolved_device
+
+        # Move graphs to resolved device.
         if data.size > 0:
             g = data.flat[0]
-            if getattr(g, "x", None) is not None:
-                inferred = cast(torch.Tensor, g.x).device
-        resolved_device = (
-            device
-            if device is not None
-            else (inferred if inferred is not None else torch.empty(0).device)
-        )
-        self._device = resolved_device
-        # Move graphs to resolved device
-        if data.size > 0:
-            for graph in self.data.flat:
-                graph.to(str(resolved_device))
+            assert isinstance(g.x, torch.Tensor)
+            if g.x.device != resolved_device:
+                for graph in self.data.flat:
+                    graph.to(str(resolved_device))
+
+    @property
+    def device(self) -> torch.device:
+        """The device on which the states are stored.
+
+        Returns:
+            The device of the underlying array of GeometricData.
+        """
+        assert self._device is not None
+        return self._device
+
+    def to(self, device: torch.device) -> GraphStates:
+        """Moves the GraphStates to the specified device.
+
+        Args:
+            device: The device to move to.
+
+        Returns:
+            The GraphStates object on the specified device.
+        """
+        for graph in self.data.flat:
+            graph.to(str(device))
+        self._device = device
+
+        return self
 
     @property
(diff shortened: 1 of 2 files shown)


src/gfn/states.py
```diff
@@ -715,13 +755,22 @@ def __init__(
...
+        # Resolve device per instance: prefer explicit, else infer, else default
+        inferred: torch.device | None = None
         if data.size > 0:
             g = data.flat[0]
-            if self._device is None:
-                self._device = cast(torch.Tensor, g.x).device
-            else:
-                ensure_same_device(self._device, cast(torch.Tensor, g.x).device)
+            if getattr(g, "x", None) is not None:
+                inferred = cast(torch.Tensor, g.x).device
```
hyeok9855: I think it is better to use the assertion that g has x.

src/gfn/states.py
```diff
@@ -715,13 +755,22 @@ def __init__(
         self.categorical_node_features = categorical_node_features
         self.categorical_edge_features = categorical_edge_features
         self.data = data
-        self._device = device
+        # Resolve device per instance: prefer explicit, else infer, else default
+        inferred: torch.device | None = None
```
hyeok9855: I would prefer `inferred_device`

src/gfn/states.py
```diff
@@ -715,13 +755,22 @@ def __init__(
...
+        resolved_device = (
+            device
+            if device is not None
+            else (inferred if inferred is not None else torch.empty(0).device)
+        )
+        self._device = resolved_device
+        # Move graphs to resolved device
+        if data.size > 0:
+            for graph in self.data.flat:
+                graph.to(str(resolved_device))
```
hyeok9855: Enumerating all the graphs everytime we initialise the state can be inefficient. Maybe we want to do this only when the first graph has different device with `self._device` (or `resolved_device), just assuming all graphs in the array have the same device.

src/gfn/states.py
```diff
@@ -1235,7 +1286,17 @@ def is_sink_state(self) -> torch.Tensor:
...
-        return self._compare(self.sf)
+        g = self.sf
+        if getattr(g, "x", None) is not None:
+            try:
+                ensure_same_device(self.device, cast(torch.Tensor, g.x).device)  # type: ignore[attr-defined]
+                other = g
+            except ValueError:
+                other = g.clone()
+                other.to(str(self.device))
+        else:
```
hyeok9855: same here, we might not need `ensure_same_device` and `.to` would be enough

src/gfn/states.py
```diff
@@ -414,6 +424,18 @@ def stack(cls, states: Sequence[States]) -> States:
...
+        """Moves the States tensor to the specified device in-place.
+
+        Args:
+            device: The device to move to.
+
+        Returns:
+            The States object on the specified device.
+        """
+        self.tensor = self.tensor.to(device)
+        return self
```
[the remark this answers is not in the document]
younik: also, I think we shold update self.device as well

testing/test_states.py
```diff
@@ -772,3 +780,212 @@ def test_graph_masks_device_on_cuda(datas):
...
     _assert_tensordict_on_device(stacked.backward_masks, stacked.device)
+
+
+def _cpu_devices():
+    return torch.device("cpu"), torch.device("cpu")
+
+
+def _cpu_and_meta_devices():
+    # Meta is a dummy device for mismatch-only tests; do not transfer back from meta
+    return torch.device("cpu"), torch.device("meta")
```
younik: I think removing these functions (and write the one line implementation instead) will make tests clearer

testing/conftest.py
```diff
@@ -6,3 +6,12 @@
...
     torch.set_default_device("cpu")
+
+
+# Fixture to reset torch deterministic algorithms state after each test.
+@pytest.fixture(autouse=True)
+def reset_torch_determinism():
+    """Ensure deterministic algorithms don't leak between tests"""
+    original_state = torch.are_deterministic_algorithms_enabled()
+    yield
+    torch.use_deterministic_algorithms(original_state)
```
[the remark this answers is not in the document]
josephdviviano: Determinism persists across tests if I don't do this. Seeding sets determinism

src/gfn/states.py
josephdviviano: I'm pretty sure I did this ensure same device stuff for a good reason actually. I'm going to keep it instead of deleting it for no reason.

josephdviviano: we had a bug where multiple states shared memory and it was messing up training. i dont have the bandwidth to look into this more.

testing/test_states.py
josephdviviano: makes sense!

PR #436, opened 2025-11-24
Container tests

hyeok9855:

## Description
 

This PR improves the tests for the containers (`Transitions` and `StatesContainer`).

Before this PR, I found that the transitions in `testing/test_containers.py` are mostly invalid, making it a bit tricky to add new tests on top of them.

This PR uses the trajectory sampling in `testing/test_samplers_and_containers.py` (renamed from `testing/test_samplers_and_trajectories.py`) to obtain `Transitions` and `StatesContainer`, rather than manually defining all components.

The tests for containers moved from `testing/test_containers.py` to `testing/test_samplers_and_containers.py`, considering their close relevance to trajectory sampling. 

The test coverage is the same as before (in terms of the functionalities tested). However, the tests are now run across multiple environments (whereas before this PR it was only run in a single env, DiscreteEBM).

### DIFF
--- a/testing/test_containers.py
+++ b/testing/test_containers.py
@@ -1,0 +1,240 @@
+# import pytest
+# import torch
+
+# from gfn.containers import StatesContainer, Trajectories, Transitions
+# from gfn.env import Env
+# from gfn.gym.discrete_ebm import DiscreteEBM
+
+# TODO: these transitions are actually invalid under the envs provided. We
+#       should fix that (probably by making a dummy environment). This is kept
+#       now as a placeholder.
+
+# def transitions_containers(env: Env):
+#     """Creates Transitions containers."""
+
+#     # Create some states
+#     states1 = env.states_from_tensor(torch.tensor([[0, 0], [1, 1]]))
+#     actions1 = env.actions_from_tensor(torch.tensor([[1], [0]]))
+#     is_terminating1 = torch.tensor([False, True])
+#     next_states1 = env.states_from_tensor(torch.tensor([[1, 0], [1, 1]]))
+#     log_probs1 = torch.tensor([-0.5, -0.3])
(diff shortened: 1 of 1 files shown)


josephdviviano:
Hi @hyeok9855 I'm a bit confused here - if these transitions were invalid, why where they passing for so long? These particular tests have been around for a long time (since March!)

https://github.com/GFNOrg/torchgfn/commit/4e5746661aa8b30646f995bb335be523103a2190

While I like what you've added I'm hesitant to throw away these other tests. Can you explain why the transitions are invalid?

hyeok9855:
@josephdviviano 

Before, in `testing/test_containers.py`, we manually assign the tensors to `States` and `Actions` to make containers, ignoring the environment.

e.g., [here](https://github.com/GFNOrg/torchgfn/blob/c3f309694b24aa4fd3cfae720e40c0a11eecb968/testing/test_containers.py#L31), we assign `torch.tensor([[2, 1], [0, 2]])` to the `States`, which is not compatible with the given env, which is `DiscreteEBM` (since its states can only have a value in `{-1, 0, 1}`).

josephdviviano:
I understand! Can we _comment out_ these tests instead of removing them entirely, to be fixed in a follow up PR? Just leave a TODO stating that the states and actions do no respect the environment.

I like having tests where the full contents can be easily verified by eye, but agree 1) this should be fixed to match the env and 2) it's not crucial this is fixed right away.

hyeok9855:
@josephdviviano 

I think an LLM can do the job (making valid transitions) quite easily, if you want to keep them. However, it will make the test code quite bulky if we want to test across multiple environments. Also, if an env changes, we might need to change the assigned values manually...

So, I think it is okay to remove the legacy, unless the new test loses the test coverage.

PR #441, opened 2025-11-28
Jit test DO NOT MERGE 

josephdviviano:
**DON'T MERGE THIS - IT'S GARBAGE**

## Description
Left here to share some results with the team.

The amount of disturbance compiling the sampling path requires is quite daunting, and I think would make this library significantly harder to use.

However, some small changes here might be useful addition to the library, so I'll add them in seperate PRs.


josephdviviano:
I'm closing this because the main ideas from it (that are any good) are slowly getting integrated by me into other PRs, and this one is really ugly.

PR #444, opened 2025-11-30
changed performance mode to deterministic mode -- which is off by def…

josephdviviano:

## Description
changed performance mode to deterministic mode -- which is off by default, so we can make more effective use of compute by default and only run deterministically when explicitly requested by the user

### DIFF
--- a/src/gfn/utils/common.py
+++ b/src/gfn/utils/common.py
@@ -313,8 +313,14 @@ def temporarily_set_seed(seed):
 
 def make_dataloader_seed_fns(
     base_seed: int,
+    deterministic_mode: bool = False,
 ) -> Tuple[Callable[[int], None], torch.Generator]:
     """Return `(worker_init_fn, generator)` for DataLoader reproducibility.
+
+    Args:
+        base_seed: The base seed to use for the DataLoader.
+        deterministic_mode: If True, uses deterministic behavior for better
+            reproducibility at the cost of performance.
 
     Example
     -------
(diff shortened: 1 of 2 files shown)


src/gfn/utils/common.py
```diff
@@ -186,19 +186,22 @@ def set_seed(seed: int, performance_mode: bool = False) -> None:
...
-            else:
+            if deterministic_mode:
                 # For reproducibility in GPU training, we still want deterministic
                 # CPU operations
                 os.environ["OMP_NUM_THREADS"] = "1"
                 os.environ["MKL_NUM_THREADS"] = "1"
+            else:
+                # For GPU training, we can use multiple threads for CPU operations
+                os.environ["OMP_NUM_THREADS"] = str(num_cpus)
+                os.environ["MKL_NUM_THREADS"] = str(num_cpus)
```
younik: isn't this the default behavior? If yes, I think we can remove this else branch

src/gfn/utils/common.py
```diff
@@ -332,7 +331,8 @@ def make_dataloader_seed_fns(
 
     def _worker_init_fn(worker_id: int) -> None:  # pragma: no cover
         # Each worker gets a distinct seed in the same pattern used for ranks.
-        set_seed(base_seed + worker_id, performance_mode=False)
+        # TODO: Can this be false?
+        set_seed(base_seed + worker_id, deterministic_mode=True)
```
younik: I think it should be false by default; as it is the default torch behavior

younik (approved):

PR #445, opened 2025-12-05
Place conditions within States

hyeok9855:

## Description
This PR tries to relocate the conditions from elsewhere to inside the `States`. There're pros and cons of this design.

### Pros
- Cleaner API, e.g., we can simply use `env.reward(states)` instead of `env.reward(states, conditions)`
- Trajectory management becomes easier since conditions are handled in `States`

### Cons
- Redundancy; the shape of condition tensor for batch of trajectories is `(max_length, batch_size, condition_dim)`, i.e., we now need to store `max_length` copy of conditions (possibly increase memory overhead)
- `States` is now a bit messy
- Less intuitive (personally); conditions can exist without `States`

### DIFF
--- a/src/gfn/env.py
+++ b/src/gfn/env.py
@@ -491,17 +491,6 @@ class ConditionalEnv(Env, ABC):
             batch_shape=batch_shape, random=random, sink=sink, conditions=conditions
         )
 
-    def reward(self, states: States) -> torch.Tensor:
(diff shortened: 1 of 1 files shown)


younik:
> I actually like this. I'm curious to hear @younik 's feedback on this design.

I also prefer this design, it looks cleaner, even if as far as I understood the conditioning cannot change during an episode.

PR #509, opened 2026-04-06
Coordinator rank, async scoring, and replay buffer refactor

josephdviviano:
## Summary

This PR adds coordinator rank support, async score collection, and refactors the replay buffer hierarchy to fix subclass bypass bugs. It also simplifies `train_hypergrid.py` by removing distributed/multinode code that has moved elsewhere.

### Coordinator rank (`src/gfn/utils/distributed.py`)
- Add `coordinator_rank` field and `is_coordinator_rank()` to `DistributedContext`.
- `initialize_distributed_compute()` accepts `use_coordinator`: when enabled, the last rank is reserved as a coordinator that aggregates mode discoveries across buffer managers.
- `initialize_distributed_compute_mpi4py()` accepts `num_coordinators` and adjusts `num_training_ranks` accordingly.
- Added assertion that `num_training_ranks > 0` after subtracting buffers and coordinators.

### Sender rank in scoring (`src/gfn/containers/replay_buffer_manager.py`)
- `ReplayBufferManager.run()` now passes `sender_rank` to the scoring function, enabling per-rank discovery attribution and discovery-rate computation.
- `default_scoring_function` signature updated to accept `sender_rank`.
- Removed unused `ContainerUnion` import (moved to replay_buffer).

### New message types (`src/gfn/containers/message.py`)
- Added `MODE_REPORT` (buffer -> coordinator: new mode hashes + training rank) and `GET_POPULATION_STATS` (training rank -> coordinator: request stats) to `MessageType`.
- Replaced `torch.ByteTensor(list(...))` with `torch.frombuffer(bytearray(...), dtype=torch.uint8)` to avoid 28x memory amplification (F-RB-9).
- Replaced `byte_tensor.tolist()` with `byte_tensor.numpy()` in deserialization for efficiency.
- Added `# type: ignore[return-value]` to satisfy pyright (`torch.frombuffer` returns `Tensor` at the type level).

### Async score collection (`src/gfn/containers/replay_buffer.py`)
- New `async_score` parameter: when enabled, trajectory sends are fire-and-forget; scores are collected lazily on the next `add()` call (1-iteration stale), decoupling training throughput from buffer scoring latency.
- New methods: `_send_objs_async()`, `_collect_pending_score()`, `drain_pending_score()` (with 30s timeout for graceful shutdown).
- Split `_send_objs()` into `_send_data()` + `_recv_score()` for reuse by both sync and async paths.

### Replay buffer subclass fix (F-RB-1)
- Renamed `_add_objs()` → `_local_add()` as the override hook for local insertion logic.
- `NormBasedDiversePrioritizedReplayBuffer` and `TerminatingStateBuffer` now override `_local_add()` instead of `add()`, so remote buffer communication in the base class is no longer bypassed.
- Flattened deeply nested if/else in `NormBasedDiversePrioritizedReplayBuffer._local_add()` for readability.

### Pyright fixes
- `env.py`: Use public `.conditions` property instead of protected `._conditions` in `_step()` and `_backward_step()`.
- `message.py`: Add type-ignore comment for `torch.frombuffer` return type.

### Simplification of `train_hypergrid.py`
- Script is now a clean **single-process** training example as intended.

### Misc
- `.pre-commit-config.yaml`: Exclude `benchmark/` directory from pre-commit hooks.
- `test_scripts.py`: Catch `OSError` instead of `FileNotFoundError` for `plc_wrapper_main` availability check.

## Test plan
- [x] `pre-commit run --all-files` passes (autoflake, isort, black, flake8, pyright, pytest)
- [ ] Verify distributed training with `async_score=True` and `use_coordinator=True`
- [ ] Verify `drain_pending_score()` handles buffer manager crash gracefully


chirayuharyan (approved):
Looks good to me.

josephdviviano:
@chirayuharyan thanks - I will have a few more changes for this.

josephdviviano:
@chirayuharyan -- can you have a second look? I'm going to merge it in so we can move fast (we can always roll it back or patch issues you uncover). See the updated PR message for details.

PR #510, opened 2026-04-09
Add backend-agnostic collective wrappers

josephdviviano:
## Summary
- Adds `get_world_size`, `all_reduce`, `all_gather`, and `broadcast` backend-agnostic wrappers to `gfn.utils.distributed`, complementing the `send`/`recv`/`barrier`/`get_rank` wrappers from #489.
- Adds `group` parameter support to existing `get_rank` wrapper.
- All wrappers support both `"torch"` and `"mpi"` backends with the same API.

## Motivation
PR #489 added point-to-point wrappers, but downstream code (gfn-communities) also uses collective operations (`all_reduce` for gradient sync, `all_gather` for flag/metric exchange, `broadcast` for weight sync). These wrappers enable full migration from `torch.distributed` to mpi4py without maintaining parallel implementations.

## Test plan
- [x] Existing `test_utils.py` and `test_dependency_free.py` pass
- [x] All wrappers import correctly
- [ ] Integration testing with mpi4py on cluster (gfn-communities migration)

### DIFF
--- a/testing/test_distributed_wrappers.py
+++ b/testing/test_distributed_wrappers.py
@@ -1,0 +1,142 @@
+"""Tests for backend-agnostic collective wrappers in gfn.utils.distributed.
+
+These tests run in a single process without initializing any distributed
+backend, so they can only exercise:
+  - error paths (unknown backend)
+  - the torch backend via a minimal single-process process group
+  - basic API contracts (return types, shapes)
+
+Full multi-process / multi-rank tests require MPI or torch.distributed
+launchers and are not feasible in standard CI.
+"""
+
+import pytest
+import torch
+import torch.distributed as dist
+
+from gfn.utils.distributed import (
+    all_gather,
+    all_reduce,
+    barrier,
+    broadcast,
+    get_rank,
+    get_world_size,
+    recv,
+    send,
+)
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture(scope="module")
+def torch_pg():
+    """Initialize a single-process torch.distributed process group for testing.
+
+    Uses the gloo backend (CPU-only, no GPU required). Torn down after the
+    module's tests complete.
+    """
+    if dist.is_initialized():
+        yield None
+        return
+
+    dist.init_process_group(
+        backend="gloo",
+        init_method="tcp://127.0.0.1:29500",
+        rank=0,
+        world_size=1,
+    )
+    yield None
+    dist.destroy_process_group()
+
+
+# ---------------------------------------------------------------------------
+# Unknown backend error tests (no distributed init needed)
+# ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


josephdviviano:
## Review follow-up (eccc74a2)

Addressed the following from code review:

- **Type annotations**: Added to all 8 wrapper functions (`send`, `recv`, `barrier`, `get_rank`, `get_world_size`, `all_reduce`, `all_gather`, `broadcast`) and to `send_termination_signal` / `get_metadata`
- **send/recv documentation**: Documented the protocol differences between backends — torch uses length-prefixed 2-message (tag=0 for length, tag=1 for payload), MPI uses a single message with `Probe` for size discovery. Noted that sender and receiver must use the same backend
- **Explicit MPI tag**: Added `tag=0` to `comm.Send` for clarity (was relying on default)
- **Device preservation**: Fixed MPI paths in `all_reduce`, `all_gather`, `broadcast` to copy results back to the original tensor's device (was silently producing CPU tensors for GPU inputs)
- **pyright fix**: Added `assert self.remote_manager_rank is not None` in `_send_data` to satisfy type narrowing
- **TerminatingStateBuffer**: Made `communication_backend` an explicit kwarg (consistent with `timing`)
- **timing_log()**: Now reports total, count, and mean per key
- **Tests**: Added 16 unit tests — 8 for unknown-backend error paths, 8 single-process gloo integration tests covering `get_rank`, `get_world_size`, `barrier`, `all_reduce` (SUM/MAX/MIN), `all_gather`, `broadcast`. Note: `send`/`recv` can't be tested in single-process gloo (segfaults on self-send); requires multi-process launcher

chirayuharyan (approved):
Looks good to me. 👍

testing/test_distributed_wrappers.py
```diff
@@ -0,0 +1,142 @@
...
+        with pytest.raises(ValueError, match="Unknown backend"):
+            broadcast(torch.zeros(1), src=0, backend="nccl_fake")
+
+
+# ---------------------------------------------------------------------------
+# Torch backend single-process tests
+# ---------------------------------------------------------------------------
+
+
+class TestTorchBackendSingleProcess:
```
chirayuharyan, line 100: It will be useful to add test for mpi4py backend as well.

chirayuharyan (changes requested):

testing/test_distributed_wrappers.py
josephdviviano: i'm adding now and will verify on our cluster

josephdviviano: done, verified (but have to be run on different hardware than available here - so so not explicitly verified by CI)

chirayuharyan: Thanks for the changes. Everything works on my hardware too. :)

chirayuharyan (approved):
LGTM

PR #515, opened 2026-04-12
Use dedicated MPI tag for metadata queries

josephdviviano:
## Summary

- Adds `tag` parameter to `send()` and `recv()` in `distributed.py` (default 0, backward compatible)
- Defines `DATA_TAG=0` and `METADATA_TAG=1` constants in `replay_buffer_manager.py`
- Metadata responses now use `METADATA_TAG` so they never collide with pending async score responses on `DATA_TAG`

**Bug:** When `async_score=True` (the default), `get_metadata()` would receive a stale score dict instead of the metadata response, because both used MPI tag 0. This caused `n_modes_found` to be silently wrong or trigger an assertion failure.

**Fix:** The metadata response channel is now fully independent of the data channel via tag separation.

For the torch backend, tags are mapped to `2*tag` / `2*tag+1` (length prefix + payload) to stay compatible with the existing two-message protocol.

## Test plan

- [x] K=4 SA diagnostic: rank 0 reports `global=22523` vs `local=7758` — correctly deduplicated
- [x] Ranks 1-3 report only `n_modes_found_local` (no global key)
- [ ] Run existing torchgfn test suite to verify backward compatibility

### DIFF
--- a/src/gfn/containers/replay_buffer_manager.py
+++ b/src/gfn/containers/replay_buffer_manager.py
@@ -137,6 +137,8 @@ class ReplayBufferManager:
         # listens on tag=0 for all incoming messages).  The *response* comes
         # back on METADATA_TAG so it cannot collide with async score replies.
         send(msg_bytes, dst_rank=manager_rank, backend=backend)
-        _src_rank, metadata_tensor = recv(manager_rank, backend=backend, tag=METADATA_TAG)
+        _src_rank, metadata_tensor = recv(
+            manager_rank, backend=backend, tag=METADATA_TAG
+        )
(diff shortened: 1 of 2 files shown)


josephdviviano:
@chirayuharyan 

This PR actually has a crucial bugfix for mode tracking that needs a double check from you.

We use a unique MPI tag for data send/recv and metadata queries (i.e., global n modes found).

josephdviviano:
It's associated with this commit : https://github.com/GFNOrg/gfn-communities/commit/13e792ad5836a42c34d4943380901f90081d4671

chirayuharyan (approved):
LGTM :)

PR #523, opened 2026-04-28
Lazy sorting added for replay buffer

chirayuharyan:

### Summary

Adds an opt-in `lazy_sort` mode to `ReplayBuffer` that defers concatenation and sorting until the buffer reaches 2x capacity, significantly reducing the cost of `_local_add` on buffer manager ranks.

### Problem

On buffer manager ranks, every incoming batch triggered an `extend()` (which copies the entire buffer via `torch.cat`) followed by an `argsort` + truncation. With 32 training agents each submitting ~100 batches, this resulted in ~3,200 extend-sort-truncate cycles — dominated by the quadratic copy cost of repeatedly concatenating onto a growing tensor.

### Changes

**replay_buffer.py**

- **`lazy_sort` parameter** — new `bool` flag on `ReplayBuffer.__init__` (default `False`, backward compatible). When enabled, incoming batches are accumulated in `_pending_batches` list instead of being immediately concatenated.
- **`_flush_pending()`** — merges all pending batches into a single combined batch first, then extends `training_container` once — avoids the quadratic cost of extending one-by-one.
- **`_sort_and_truncate()`** — extracted from `_local_add` into a helper; argsorts by `log_rewards` (if prioritized) and truncates to capacity. Both paths (lazy and eager) share this method.
- **`_local_add()`** — branches on `self.lazy_sort`: lazy path accumulates and flushes at 2x capacity; eager path preserves the original extend-sort-truncate-every-call behavior.
- **`__len__`** — includes `_pending_len` so length queries reflect uncommitted items.
- **`sample()`** — samples from committed container only (does not flush pending). Guard updated to handle empty `training_container`.
- **`save()`** — calls `_flush_pending()` before serialization.
- **`NormBasedDiversePrioritizedReplayBuffer`** — `lazy_sort` parameter threaded through to `super().__init__`. `_local_add` flushes pending batches when the buffer is full (diversity filtering requires a consistent view), but delegates to the base lazy path while filling up.

**replay_buffer_manager.py**

- Passes `lazy_sort=True` and `timing=self.timing` to both `ReplayBuffer` and `NormBasedDiversePrioritizedReplayBuffer` constructors.

### Performance

Benchmarked on 32 training agents, 1 buffer manager (Intel EMR (64 cores) x 2, single node), 100 iterations:

| Phase | Before (eager) | After (lazy sort) | Reduction |
|---|---|---|---|
| **replay_add** (total) | 173.6s (3200 calls) | 97.6s (3200 calls) | **44%** |
| local_add/extend | 53.2s (3200 calls) | 61.7s (399 calls) | per-call larger, but fewer calls |
| local_add/sort | 111.6s (3200 calls) | 33.7s (399 calls) | **70%** |
| local_add/truncate | 0.3s (3200 calls) | 0.06s (399 calls) | — |

### Backward Compatibility

`lazy_sort` defaults to `False` — existing code is unaffected. The eager path is unchanged aside from being wrapped in timers and delegating to the shared `_sort_and_truncate` helper.

### DIFF
--- a/src/gfn/containers/replay_buffer.py
+++ b/src/gfn/containers/replay_buffer.py
@@ -76,6 +76,7 @@ class ReplayBuffer:
         timing: bool = False,
         async_score: bool = False,
         async_comm: bool = False,
+        lazy_sort: bool = False,
     ):
         """Initializes a ReplayBuffer instance.
 
@@ -101,6 +102,9 @@ class ReplayBuffer:
                 submission (deferred score collection on next ``add()``).
                 The agent blocks only to collect the previous score and
                 to wait for the prior ``isend`` to complete.
+            lazy_sort: If True, defer concatenation and sorting until the
+                buffer reaches 2x capacity. Useful for buffer manager ranks
+                that only accumulate data and don't sample every iteration.
         """
         self.env = env
         self.capacity = capacity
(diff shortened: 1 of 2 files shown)


josephdviviano (approved):
lgtm, great work, thank you

PR #524, opened 2026-05-07
Add baseline filtering and scoring options to ReplayBuffer

chirayuharyan:

### TLDR
- Added baseline_filtering to send only trajectories whose log-reward exceeds the current remote buffer baseline.
- Added scoring_only mode to send only terminating states and log-rewards instead of full trajectory data when trajectory storage is unnecessary.

### More Details
This pull request introduces two major new features to the replay buffer system: baseline filtering and scoring-only communication. These changes allow agents to avoid sending low-value trajectories and to reduce communication costs by sending only essential data when appropriate. The buffer manager now also supports operation without local storage, and tracks/reporting for these new modes has been added.

**Replay Buffer Improvements:**

* Added `baseline_filtering` option to `ReplayBuffer` to filter out trajectories with log-reward below the current baseline (worst item in the remote buffer); includes logic for updating and applying the baseline, and tracking statistics on filtered/skipped sends. [[1]](diffhunk://#diff-34d32d1df13b0243e92f09447b23b3ef52cba940af4ac1e399585b9bd210721aR80-R81) [[2]](diffhunk://#diff-34d32d1df13b0243e92f09447b23b3ef52cba940af4ac1e399585b9bd210721aL107-R117) [[3]](diffhunk://#diff-34d32d1df13b0243e92f09447b23b3ef52cba940af4ac1e399585b9bd210721aR148-R159) [[4]](diffhunk://#diff-34d32d1df13b0243e92f09447b23b3ef52cba940af4ac1e399585b9bd210721aR181-R184) [[5]](diffhunk://#diff-34d32d1df13b0243e92f09447b23b3ef52cba940af4ac1e399585b9bd210721aL194-R255) [[6]](diffhunk://#diff-34d32d1df13b0243e92f09447b23b3ef52cba940af4ac1e399585b9bd210721aR280-R356) [[7]](diffhunk://#diff-34d32d1df13b0243e92f09447b23b3ef52cba940af4ac1e399585b9bd210721aR676-R684)
* Added `scoring_only` option to `ReplayBuffer`, allowing agents to send only terminating states and log-rewards (not full trajectories) to the remote buffer manager, reducing communication and serialization overhead. [[1]](diffhunk://#diff-34d32d1df13b0243e92f09447b23b3ef52cba940af4ac1e399585b9bd210721aR80-R81) [[2]](diffhunk://#diff-34d32d1df13b0243e92f09447b23b3ef52cba940af4ac1e399585b9bd210721aL107-R117) [[3]](diffhunk://#diff-34d32d1df13b0243e92f09447b23b3ef52cba940af4ac1e399585b9bd210721aR148-R159) [[4]](diffhunk://#diff-34d32d1df13b0243e92f09447b23b3ef52cba940af4ac1e399585b9bd210721aR280-R356)

**Replay Buffer Manager Enhancements:**

* Added `store_locally` option to `ReplayBufferManager` to allow operation without storing data locally; data is only scored and not retained if this is set to `False`. [[1]](diffhunk://#diff-71e4ad6e858554524fc404374539841fde0d96bd2a094af2ba7d215d89b96facR35-R37) [[2]](diffhunk://#diff-71e4ad6e858554524fc404374539841fde0d96bd2a094af2ba7d215d89b96facR124-R131) [[3]](diffhunk://#diff-71e4ad6e858554524fc404374539841fde0d96bd2a094af2ba7d215d89b96facR186-R193)
* Buffer manager now injects a `baseline_log_reward` field into score responses (when full) to communicate the current baseline to agents using baseline filtering. [[1]](diffhunk://#diff-71e4ad6e858554524fc404374539841fde0d96bd2a094af2ba7d215d89b96facR76-R94) [[2]](diffhunk://#diff-71e4ad6e858554524fc404374539841fde0d96bd2a094af2ba7d215d89b96facR124-R131) [[3]](diffhunk://#diff-71e4ad6e858554524fc404374539841fde0d96bd2a094af2ba7d215d89b96facR186-R193)

**Monitoring and Logging:**

* Enhanced timing logs in `ReplayBuffer` to report statistics on baseline filtering: proportion of items kept/filtered and number of sends skipped due to filtering.

### DIFF
--- a/src/gfn/containers/replay_buffer.py
+++ b/src/gfn/containers/replay_buffer.py
@@ -77,6 +77,8 @@ class ReplayBuffer:
         async_score: bool = False,
         async_comm: bool = False,
         lazy_sort: bool = False,
+        baseline_filtering: bool = False,
+        scoring_only: bool = False,
     ):
         """Initializes a ReplayBuffer instance.
 
@@ -104,8 +106,15 @@ class ReplayBuffer:
                 to wait for the prior ``isend`` to complete.
             lazy_sort: If True, defer concatenation and sorting until the
                 buffer reaches 2x capacity. Useful for buffer manager ranks
-                that only accumulate data and don't sample every iteration.
-        """
+                that only accumulate data and don't sample every iteration.            baseline_filtering: If True, filter trajectories before sending
+                to the remote buffer manager. Only trajectories with
+                log-reward above the baseline (worst trajectory in the
+                remote buffer) are sent. The baseline is communicated
+                back by the buffer manager in the score response.
+            scoring_only: If True, only send terminating states and
+                log-rewards to the remote buffer manager (not full
+                trajectories). Use when the manager does not store data
+                locally and only needs scoring information."""
         self.env = env
         self.capacity = capacity
         self._is_full = False
@@ -136,6 +145,18 @@ class ReplayBuffer:
         self._pending_batches: list[ContainerUnion] = []
         self._pending_len: int = 0
 
+        # Baseline filtering: skip sending trajectories worse than the
+        # remote buffer's worst item. Updated from score responses.
+        self.baseline_filtering = baseline_filtering
+        self._baseline_log_reward: float = float("-inf")
+        # Filtering counters (tracked when timing=True).
+        self._baseline_total: int = 0
+        self._baseline_kept: int = 0
+        self._baseline_skipped_sends: int = 0
+
+        # Scoring-only: send only terminating states + log_rewards.
+        self.scoring_only = scoring_only
+
     @property
     def device(self) -> torch.device:
         """The device on which the buffer's data is stored.
@@ -156,6 +177,10 @@ class ReplayBuffer:
         call returns None (no pending score yet), and subsequent calls return the
         score from the *previous* submission.  This decouples training throughput
         from buffer scoring latency.
+
+        When ``baseline_filtering`` is enabled, only trajectories with log-reward
+        above the remote buffer's baseline are sent.  If all trajectories in the
+        pending batch are below the baseline, the send is skipped entirely.
 
         Args:
             training_container: The Trajectories, Transitions, or StatesContainer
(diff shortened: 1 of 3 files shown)


src/gfn/containers/replay_buffer.py
```diff
@@ -236,6 +277,83 @@ def _collect_pending_score(self) -> dict[str, float] | None:
...
+        log_rewards = container.log_rewards
+        n = len(terminating_states)
+        return StatesContainer(
+            env=self.env,
+            states=terminating_states,
+            is_terminating=torch.ones(
+                n, dtype=torch.bool, device=terminating_states.device
+            ),
+            log_rewards=log_rewards,
+        )
```
younik, line 387: I suspect this doesn't work with full Trajectories. The script we are running are using terminal states as container, thus log_rewards refer to them, but in general, if container are trajectories, container.terminating_states and container.log_rewards have different shapes

src/gfn/containers/replay_buffer.py
```diff
@@ -236,6 +277,83 @@ def _collect_pending_score(self) -> dict[str, float] | None:
...
+
+        if mask.all():
+            return container  # All above baseline — send everything.
+        if not mask.any():
+            if self.timing:
+                self._baseline_skipped_sends += 1
+            return None  # All below baseline — skip send.
+
+        indices = torch.where(mask)[0]
+        return container[indices]
```
younik, line 353: if we want to have this for general-use, we should keep the container type consistent. 

Think about what happen with the current code if container is a trajectory, it seems to me we select only the states with reward higher to baseline, but I believe we should return the full trajectories based on a baseline constraint (e.g. sum of rewards in the trajectory). 
Maybe `container[indices]` even crashes for trajectories because bad indexing, not sure

younik (commented):
Thanks for this!

My only concern is that this code is a bit overfitted to our usecase, but we are merging to broader torchgfn.
I would land these changes anyway, but we need at least a full test coverage for all the Container types.

chirayuharyan:
Thank you Joseph for implementing these changes; the updates look great and really enhance the feature. Good to go on my end.
And thanks, Omar, for the thorough review and helpful feedback. You made a great catch regarding the potential errors with different Container Types versus Trajectories, and Joseph's changes resolve it nicely.

younik (approved):
Thanks!

